PR #1, opened 2016-11-27
Sync timers

wzli:
Added function to sync timers to arbitrary count offset.
Added synchronization of timers 1 and 8 with 90 degrees offset to initialization.


madcowswe (changes requested):
Can you please make it so that MOE is set to its original state, rather than hard disable, hard enable?
Thx!

madcowswe (approved):

PR #3, opened 2017-01-25
Support debugging from the command line

fyquah:


.gdbinit
```diff
@@ -0,0 +1,10 @@
+break Src/main.c:89
```
madcowswe, line 1: Is it possible to tag a function name or symbol to be the breakpoint? A line number may not be stable.

Makefile
```diff
@@ -150,6 +150,16 @@ $(BUILD_DIR)/%.bin: $(BUILD_DIR)/%.elf | $(BUILD_DIR)
 $(BUILD_DIR):
 	mkdir -p $@
 
+
+debug: build/ODriveFirmware.elf
+	{ openocd -c gdb_port 3333 -c telnet_port 4444 -f interface/stlink-v2.cfg -f target/stm32f4x_stlink.cfg -c 'echo "Started by GNU ARM Eclipse"' & } && echo $$!>.openocd.pid
+	arm-none-eabi-gdb -ix .gdbinit $<; cat .openocd.pid | xargs kill -2; rm .openocd.pid
```
madcowswe, line 156: Can you add some comments about what exactly the second half of this command does?

Makefile
```diff
@@ -150,6 +150,16 @@ $(BUILD_DIR)/%.bin: $(BUILD_DIR)/%.elf | $(BUILD_DIR)
 $(BUILD_DIR):
 	mkdir -p $@
 
+
+debug: build/ODriveFirmware.elf
+	{ openocd -c gdb_port 3333 -c telnet_port 4444 -f interface/stlink-v2.cfg -f target/stm32f4x_stlink.cfg -c 'echo "Started by GNU ARM Eclipse"' & } && echo $$!>.openocd.pid
```
madcowswe, line 155: It is not started by Eclipse?

Makefile
madcowswe: Actually, this way of launching two processes and getting the pid seems overcomplicated. At work we use this command:
`arm-none-eabi-gdb ${NAME}.elf -x openocd.gdbinit`
where openocd.gdbinit contains:
```
target remote | openocd -f "interface/stlink-v2.cfg" -f "target/stm32f4x_stlink.cfg" -c "gdb_port pipe; log_output openocd.log"
monitor reset halt
```

See if you can get something like that to work?

madcowswe (commented):
Have a look at the suggestion for simpler command to get gdb going.
Thanks!

madcowswe:
Closing PR as it is redundant with the implementation in bfc00d18..156b6c5c

PR #6, opened 2017-02-27
feature/usb cdc

nickaknudson:
This pull request introduces USB CDC for ODrive.

The interface is currently very simple:
```
ODrive USB Bulk Communications
---------------------------------------------------------------------
USAGE:
	POSITION_CONTROL:
		p MOTOR_NUMBER POSITION VELOCITY CURRENT
	VELOCITY_CONTROL:
		v MOTOR_NUMBER VELOCITY CURRENT
	CURRENT_CONTROL:
		c MOTOR_NUMBER CURRENT
---------------------------------------------------------------------
```

The following example uses position control to set motor 0 to 23.5 1.5 0.5:
```
p 0 23.5 1.5 0.5
```

# Future Work
- [ ] switching from CDC USB Bulk to vendor specific USB Control and implementing custom struct packing
- [ ] better python scripts


madcowswe:
Thank you for this contribution!
I had a brief read through the changes, and it looks good!
I will do some testing, and some slight restructuring in the next couple of days (mostly to do with decoding the control packet in a function specified in the motor_control file, rather than inline in the CDC_Receive_FS).

When you say "switching from CDC USB Bulk to vendor specific USB Control", what do you mean?
I had a look here http://www.beyondlogic.org/usbnutshell/usb4.shtml and I guess interrupt transfers are good in terms of latency. But you mean even more custom than these profiles?

nickaknudson:
A USB Control endpoint would let you define a custom protocol, for instance:
```
enum packet_type {
    UNKNOWN = 0,
    POSITION_CONTROL = 1,
    VELOCITY_CONTROL = 2,
    CURRENT_CONTROL = 3
};

typedef struct motor_control_packet { 
    packet_type packet_type;
    uint8_t motor_number;
    float pos_setpoint;
    float vel_feed_forward;
    float current_feed_forward;
} motor_control_packet;
```
Then the host would send a fully formed `motor_control_packet` structure to the ODrive. This makes the control protocol much more tightly coupled on host and device, but also greatly reduces error.

As is stands right now, the USB protocol uses `sscanf` to interpret a user command. This is highly error prone if the user sends a malformed packet.

madcowswe:
I agree that's a good idea. But I understand that part as "implementing custom struct packing", but I don't see how it pertains to switching from bulk transfers?

nickaknudson:
You're right, I'm probably using the wrong terminology. It would still be a
bulk transfer.

madcowswe:
I resolved the conflicts on my branch I will be testing on here:
https://github.com/madcowswe/ODriveFirmware/tree/usb-cdc

nikonikolov:
Hi, Nick, out of curiosity, what changes did you make to the code generated by STMCubeMX. Also which version of STMCubeMX did you use? I fought with this feature for over a month and could not even get the board to show up under `/dev/` on my machine. Thanks!

nickaknudson:
@nikonikolov I also struggled for some time to even get the device to show up in `lsusb`. One of the tricks was to disable `VBUS Detect` as that pin is not hooked up on ODrive board.

madcowswe:
@nickaknudson I have some more questions, that I think would be more properly discussed in IM or over a call. I invited you (with the gmail email you made commits with) to the ODrive slack if you have time to answer some questions there?

nickaknudson:
https://github.com/pidcodes/pidcodes.github.com/pull/228

Src/usbd_cdc_if.c
```diff
@@ -0,0 +1,315 @@
...
+  * @}
+  */ 
+
+/** @defgroup USBD_CDC_Private_Defines
+  * @{
+  */ 
+/* USER CODE BEGIN PRIVATE_DEFINES */
+/* Define size for the receive and transmit buffer over CDC */
+/* It's up to user to redefine and/or remove those define */
+#define APP_RX_DATA_SIZE  4
```
madcowswe, line 74: I think the CDC default receive length is 64 bytes, which would overflow the buffer, right?

Src/usbd_cdc_if.c
```diff
@@ -0,0 +1,315 @@
...
+  *         in receiving more data while previous ones are still not sent.
+  *                 
+  * @param  Buf: Buffer of data to be received
+  * @param  Len: Number of data received (in bytes)
+  * @retval Result of the operation: USBD_OK if all operations are OK else USBD_FAIL
+  */
+static int8_t CDC_Receive_FS (uint8_t* Buf, uint32_t *Len)
+{
+  /* USER CODE BEGIN 6 */
+  USBD_CDC_SetRxBuffer(&hUsbDeviceFS, &Buf[0]);
```
madcowswe, line 265: Isn't it more clear to just set it to UserRxBufferFS again? Or do we even need to update it, since it should be set to that already?

Src/usbd_cdc_if.c
```diff
@@ -0,0 +1,315 @@
...
+  USBD_CDC_ReceivePacket(&hUsbDeviceFS);
```
madcowswe, line 266: We area already in the CDC_Receive_FS callback, so does this schedule the next packet, or are we transferring to our buffer for our current transfer right here?

madcowswe (commented):
Some questions.

madcowswe:
Please pull in https://github.com/madcowswe/ODriveFirmware/tree/usb-cdc to your branch, and check the changes I made.

Src/usbd_cdc_if.c
nickaknudson: hmm those were auto generated and I don't think that I used them anywhere

Src/usbd_cdc_if.c
nickaknudson: you're right - that buffer was auto generated to be used by user application

Src/usbd_cdc_if.c
nickaknudson: looks like its used to prepare for next packet https://github.com/madcowswe/ODriveFirmware/pull/6/files/19c96fdc8d3c8bbc2f7d6db4c7f5288038f0107d#diff-0c3dcf4a3b0a9c3af7c3c1f6e21cf930R883

Src/usbd_cdc_if.c
madcowswe: It is used in CDC_Init_FS, and carried over as the Buf in CDC_Receive_FS. I upped these buffers to 64 in my branch.

Src/usbd_cdc_if.c
madcowswe: Okay, that is also what I thought. But I wonder why it's not required to ask for the first packet in CDC_Init_FS?
I guess it's not very important, if it works, it works xP

PR #8, opened 2017-03-03
Fix for compiling under case-insensitive systems like OSX

RickP:
OSX doesn't know the difference between 'freertos.h' and 'FreeRTOS.h' - so the code doesn't compile. I renamed 'freertos.h' to  'freertos_vars.h' and changed the includes - now it compiles just fine on OSX.

Seems like my editor also removed some whitespace :)


madcowswe:
Thanks for the PR. I will merge your rename.
But can you please undo the whitespace changes? I agree that the whitespace shouldn't be there, but this is autogenerated code. I'd rather avoid changing it manually, since it will show up as different if autogenerated again.

RickP:
OK, I put the spaces back in

PR #13, opened 2017-03-13
Upgrade to CubeMX v4.20.0 and FreeRTOS v9.0.0

sirusdv:


Drivers/CMSIS/Device/ST/STM32F4xx/Include/stm32f405xx.h
```diff
@@ -169,6 +169,7 @@ typedef enum
   OTG_HS_EP1_IN_IRQn          = 75,     /*!< USB OTG HS End Point 1 In global interrupt                        */
   OTG_HS_WKUP_IRQn            = 76,     /*!< USB OTG HS Wakeup through EXTI interrupt                          */
   OTG_HS_IRQn                 = 77,     /*!< USB OTG HS global interrupt                                       */
+  HASH_RNG_IRQn               = 80,     /*!< Hash and Rng global interrupt                                     */
```
madcowswe, line 172: How did this get turned on?

sirusdv: Its not actually turned on. This is the only reference to it and the specific IRQ is not handled. I assume its generated for completeness.

PR #14, opened 2017-03-13
Use "minified" CubeMX code generator. 

sirusdv:
Its still garbage spaghetti code. I suggest we move away from CodeMX generated output and start with a fresh FreeRTOS base.


madcowswe:
Thanks for the contribution! To be honest, I think that the mess is actually slightly worse overall (the mess in main.c is worse than the reduction of files we achieve).
I'm happy we tried it though, thanks!

PR #19, opened 2017-04-03
usb interface + python api + web interface added

dirk-makerhafen:


MotorControl/low_level.h
```diff
@@ -6,18 +6,37 @@
...
+#define ERROR_FOC_TIMING 11
+#define ERROR_FOC_MEASUREMENT_TIMEOUT 12
+#define ERROR_SCAN_MOTOR_TIMING 13
+#define ERROR_FOC_VOLTAGE_TIMING 14
+#define ERROR_GATEDRIVER_INVALID_GAIN 15
+#define ERROR_PWM_SRC_FAIL 16
+
+#define POSITION_CONTROL 0
+#define VELOCITY_CONTROL 1
+#define CURRENT_CONTROL 2
```
madcowswe, line 33: Why did you switch from enum to defines?

madcowswe:
Please note that this codebase uses spaces and not tabs for indentation. I would suggest to use an editor that automatically detecs which one it is.
Don't worry about fixing it though, I'm already on it. You can check my edits over on: https://github.com/madcowswe/ODriveFirmware/tree/python_interface_and_self_tests

sorki:
I've gave this a go, had to disable safety checks for phase inductance and resistance to make it run. 

Looks like sometimes it is necessary to send set_do_selftest multiple times before controller reacts.

You're also missing shebangs (#!/usr/bin/env python) in your python mains. Only generate_api.py and ODrive.py should be executable and both should start with shebang.

Src/usbd_cdc_if.c
```diff
@@ -284,14 +284,37 @@ static int8_t CDC_Receive_FS (uint8_t* Buf, uint32_t *Len)
...
+  /* old
   uint8_t result = USBD_OK;
-  /* USER CODE BEGIN 7 */ 
   USBD_CDC_HandleTypeDef *hcdc = (USBD_CDC_HandleTypeDef*)hUsbDeviceFS.pClassData;
   if (hcdc->TxState != 0){
     return USBD_BUSY;
   }
   USBD_CDC_SetTxBuffer(&hUsbDeviceFS, Buf, Len);
   result = USBD_CDC_TransmitPacket(&hUsbDeviceFS);
+  */
```
kalvdans, line 295: Please don't leave old code in the source tree, it will be saved in the git history anyway.

Src/usbd_cdc_if.c
```diff
@@ -284,14 +284,37 @@ static int8_t CDC_Receive_FS (uint8_t* Buf, uint32_t *Len)
...
+
+	USBD_CDC_HandleTypeDef   *hcdc = (USBD_CDC_HandleTypeDef*) hUsbDeviceFS.pClassData;
+	uint8_t result = USBD_OK;
+	USBD_CDC_SetTxBuffer(&hUsbDeviceFS, Buf, Len);
+
+  do
+   {
+       result |= USBD_CDC_TransmitPacket(&hUsbDeviceFS);
```
kalvdans, line 303: You can't 'or' two enum values together with a meaningful result.

Src/usbd_cdc_if.c
```diff
@@ -284,14 +284,37 @@ static int8_t CDC_Receive_FS (uint8_t* Buf, uint32_t *Len)
...
+       //TODO can we sleep here in this thread so cpu isn't busy looping?
+   }
+   while((result != USBD_OK) && (result != USBD_FAIL));
```
kalvdans, line 306: Is this complicated loop condition same as  (result == USBD_BUSY) ?

python-interface/ODrive.py
```diff
@@ -0,0 +1,282 @@
...
+                return 1
+            return 0
+
+    # raw tty functions, lowest level
+
+    def _ttysend(self, data):
+        if DEBUG:
+            print "sending '%s'" % data
+        if self._serial != None:
+            self._serial.write(data)
```
kalvdans, line 242: Should we add carriage return or newline here? Please write a small docstring explaining if it is the caller's responsibility to add the newline.

python-interface/ODrive.py
```diff
@@ -0,0 +1,282 @@
...
+        exit(1)
+
+    TTY = sys.argv[1]
+    HOST = "127.0.0.1"
+    PORT = 12342
+
+    try:
+        HOST = sys.argv[2].split(":")[0]
+        PORT = int(sys.argv[2].split(":")[1])
+    except:
```
kalvdans, line 273: Please don't silence KeyboardInterrupt here, I expect the program to die when I press CTRL-C.
Better to be explicit:

except (IndexError, ValueError):

madcowswe:
@dirk-attraktor I lost contact with you over Slack and Email. I'm trying to reach out to you here too. Please get back in touch if you can ;D

PR #27, opened 2017-09-18
Cpp refactor

Wetmelon:


madcowswe:
- [ ] Rename .h files to .hpp if they are not extern "C"
- [ ] Use include guards instead of pragma once.

kalvdans:
Can you describe or link to the benefits we get from C++ conversion? It adds 2000 lines of code.

madcowswe:
We will redo the refactor from scratch taking into account various things we learned.

Wetmelon:
@kalvdans To answer your question though, a lot of those lines of code are taken up by hardware-version specific code that only gets inserted by pre-processor, so the actual executed code isn't much longer than originally.  However, it gives us the chance to use more abstract data types, and to structure things a little more nicely in the future.  Performance shouldn't be significantly impacted, but programming ease should be much better.

kalvdans:
Thanks @Wetmelon and @madcowswe , you scared me for a moment :) Doing the hardware header file split-up separately from the language conversion would be easier to review.

PR #28, opened 2017-09-20
Test bulk timeout fix

Wetmelon:


MotorControl/low_level.c
```diff
@@ -422,6 +422,7 @@ void motor_parse_cmd(uint8_t* buffer, int len) {
         int numscan = sscanf((const char*)buffer, "p %u %f %f %f", &motor_number, &pos_setpoint, &vel_feed_forward, &current_feed_forward);
         if (numscan == 4 && motor_number < num_motors) {
             set_pos_setpoint(&motors[motor_number], pos_setpoint, vel_feed_forward, current_feed_forward);
+            printf("M%d\n", motor_number);
```
madcowswe: I don't think this printf is appropriate, there is already a print inside each of the set_pos_setpoint, that is guarded by `#ifdef DEBUG_PRINT`. This is important because you might be polling some other variable, and you don't want to get mixed up. You could add the motor number to the existing printf instead?

Wetmelon: Oops you're right.  I didn't see that at all.  It would also spam back data if you were feeding it via a script, which we don't want.  I'll just remove the printf.

Issue #45, opened 2017-10-31
Add G-Code support


madcowswe:
We wish to have G-Code parsing and motion planning for stand-alone operation. Work has been started to port these parts of GRBL:
https://github.com/Hello1024/grbl

madcowswe:
Many double literals are used in GRBL, we'd need to use `-fsingle-precision-constant`

Hello1024:
#47 Addresses this.

yhaiovyi:
May I ask why would you want a stand-alone operation with 2 motors only? Or am I missing something?

PR #47, opened 2017-11-01
Gcode Input functionality

Hello1024:
This PR makes the USB-serial connection GCODE compatible, and should let ODrive work with a large number of existing software packages.

The X and Y axis will control motor 0 and 1.

* Untested on real hardware.  There are likely tweaks and more testing required before you can simply hook this up to your 3d print software of choice...

NOT YET READY FOR MERGING - REVIEW ONLY


Firmware/MotorControl/commands.c
```diff
@@ -137,6 +138,7 @@ void motor_parse_cmd(uint8_t* buffer, int len, SerialPrintf_t response_interface
     // TODO very hacky way of terminating sscanf at end of buffer:
     // We should do some proper struct packing instead of using sscanf altogether
     buffer[len-1] = 0;
+    buffer++;
```
kalvdans, line 147: Why skipping first byte?

Firmware/MotorControl/gcode_input.c
```diff
@@ -0,0 +1,63 @@
+#include "grbl.h"
+#include "gcode_input.h"
+
+void consume_output() {
```
kalvdans, line 4: Use (void) to denote zero function parameters in C.

Firmware/MotorControl/gcode_input.c
```diff
@@ -0,0 +1,63 @@
...
+  // Gets the current block. Returns NULL if buffer empty
```
kalvdans, line 5: Codestyle: indent with four spaces

Firmware/MotorControl/gcode_input.h
```diff
@@ -0,0 +1,10 @@
+/* Define to prevent recursive inclusion -------------------------------------*/
+#ifndef __GCODE_INPUT_H
+#define __GCODE_INPUT_H
+
+#include "commands.h"
+
+// Null terminated line of gcode to be parsed and executed.   May block until gcode is executed.
+void parse_gcode_line(unsigned char* line, SerialPrintf_t response_interface);
```
kalvdans, line 8: Use 'const char *' to denote ascii.

kalvdans (commented):
Just some small remarks on the odrive part of the code.

Firmware/MotorControl/commands.c
Hello1024: The first byte is always a $ now.  Previously, we stripped that framing byte, but since Gcode doesn't have framing bytes, it made sense to pass them to this layer.

Firmware/MotorControl/gcode_input.c
Hello1024: Done

Firmware/MotorControl/gcode_input.c
Hello1024: Done.  Haven't updated the grbl code - theres a lot of it, and the style is quite different.

Firmware/MotorControl/commands.c
kalvdans: Thanks. Adding a comment about that would be good for future readers.

Firmware/MotorControl/gcode_input.c
kalvdans: Thanks. Yes I saw it, we will have to use it unchanged to be able to pull in patches from there.

Wetmelon: Don't worry about it too much.  Once it's in VSCode, right-click -> Format Document will format it to the style we're using in ODrive (Google with slight modifications).  It'll cause issues though for updates like you said.

samuelsadok:
Since this PR is now quite far behind the devel branch I think it makes not much sense to ask to update the PR. So I'm closing this now but will leave the related issue (#45) open.

If someone wants to look at this again, from what I understand about GCode it seems that we could just add a select few commands instead of the whole GRBL library. These commands could be added directly to the ASCII protocol handler which actually already implements GCode-compatible handling of checksums and comments. Also the motion planner of GRBL code is redundant with our own trapezoidal trajectory planner now.

PR #49, opened 2017-11-01
Basic Anticogging Functionality

Wetmelon:


Firmware/MotorControl/low_level.c
```diff
@@ -819,11 +837,41 @@ static bool motor_calibration(Motor_t* motor){
...
+ * http://www.roboticsproceedings.org/rss10/p42.pdf
+ * 
+ * This implementation iterates through each encoder position,
+ * waits for zero velocity & position error,
+ * then samples the current required to maintain that position.
+ * 
+ * This holding current is added as a feedforward term in the control loop.
+ */
+bool anti_cogging_calibration(Motor_t* motor) {
+    if (motor->anticogging.calib_anticogging) {
```
madcowswe: Probably also check `anticogging.cogging_map != NULL` here, since while we do check it at the communication interface, that is prone to changing: better be sure.

Firmware/MotorControl/low_level.c
```diff
@@ -819,11 +837,41 @@ static bool motor_calibration(Motor_t* motor){
...
     motor->sensorless.pll_ki = motor->encoder.pll_ki;
-    
+
     motor->calibration_ok = true;
     return true;
 }
 
+/*
+ * This anti-Cogging calibration algorithm is based on the work at
+ * http://www.roboticsproceedings.org/rss10/p42.pdf
```
madcowswe: This is not really true, we use neither of their algorithms. And just sampling the integrator we came up with ourselves.

Firmware/MotorControl/low_level.c
```diff
@@ -819,11 +837,41 @@ static bool motor_calibration(Motor_t* motor){
...
+            return false;
+        } else {
+            motor->anticogging.index = 0;
+            set_pos_setpoint(motor, 0.0f, 0.0f, 0.0f);  // Send the motor home
+            motor->anticogging.use_anticogging = true;  // We're good to go, enable anti-cogging
+            motor->anticogging.calib_anticogging = false;
+            return true;
+        }
+    }
+    return false;
```
madcowswe, line 870: This is superfluous, and I don't think the compiler will warn you since all conditional branches return themselves.

Firmware/MotorControl/low_level.c
```diff
@@ -1224,6 +1273,18 @@ static void control_motor_loop(Motor_t* motor) {
...
         float Iq = motor->current_setpoint;
+
+        // Anti-cogging is enabled after calibration
+        // We get the current position and apply a current feed-forward
+        // ensuring that we handle negative encoder positions properly (-1 == ENCODER_CPR - 1)
+        if(motor->anticogging.use_anticogging){
+            int32_t cogPos = (int32_t)motor->encoder.pll_pos % (int32_t)ENCODER_CPR;
+            if(cogPos < 0){
+                cogPos += ENCODER_CPR;
+            }
```
madcowswe: We use the modulus function in other places, like
`int corrected_enc = encoder->encoder_state % ENCODER_CPR;` in update_rotor.
Actually there it's broken (the intention was mod, same as with your correction, but just used  the `%` remainder operator), and only saved by chance by wrap_pm_pi.

You can make a `int mod(int dividend, int divisor)` function in util.c

kalvdans: I went through the corner cases in this calculation in my head and verified it should work, provided we don't overflow int32.

madcowswe (changes requested):

kalvdans (approved):

Firmware/MotorControl/low_level.c
Wetmelon: Added.

Firmware/MotorControl/low_level.c
Wetmelon: I agree, but the compiler was throwing warnings that it reached the end of the function w/o returning so I added the `return false`

madcowswe (approved):

madcowswe:
Can you please update Firmware/CHANGELOG.md, then I'll merge it.

Issue #17, opened 2017-03-27
Clean up and robustify USB communication


madcowswe:
It should be able to be configured and commissioned over USB, and it should be able to take runtime commands from USB.

If that means that we have a G-Code parser on board, or if we hook it up through some python on the PC that then speaks some binary protocol is yet to be determined. Feel free to add suggestions.

dirk-makerhafen:
I added a simple get/set usb interface + a python code generator to create an api + a javascript webinterface with live plotting. 

I also added some code for monitoring/live plotting: my idea was this: because there is alway a compromise between "how many values" and "how often" , i added  monitoring_slots, 20 for now. The usb-api exposes a function like "monitor value xy on slot 1". If you then call "o(utput) 2" you get the first 2 monitored values. So a client can configure it to have important values on the first few slots, and read these with kHz, and less important values later in the array, so maybe every once a second it can call "o 20" to get all monitored values. The ODrive.py code makes use of these functions and automatically assigns and reads monitoring slots if you choose to monitor some value with some frequency. 

Pull request: https://github.com/madcowswe/ODriveFirmware/pull/19

alexisdal:
iwth master commit b0e57908ce5612043f9de9885dc72b257ac80cb9
i managed to get USB interface unresponsive with motor not moving anymore by using the following script

```
#!/bin/bash
while(true); do
   num_points="9999"
   echo -en "p 0 $num_points 0 0\n" > /dev/ttyACM0
   sleep 3.0
   echo -en "p 0 -$num_points 0 0\n" > /dev/ttyACM0
   sleep 3.0
done
```

(basically going back and forth)

i get the usb interface to stop responding after about 1 or 2 minutes

alexisdal:
i think it is not relevant but i can post specific motor params in low_level.c along with engine and encoder specs if that help. plz advise

madcowswe:
Okay, so for right now, here's the plan:

- We wanna keep the ASCII style comms for now.
- Main thing to improve is presenting an API that has symbols for the parameters instead of numbers
- We make all parameters use getters/setters (this is important for setting computed parameters, like control bandwith)
- We store a table of structs containing either:
  - {"name_string", TYPE_AS_ENUM, address_of_getter (or NULL), address_of_setter (or NULL)}
  - {"name_string", CONTAINER_SPECIAL_ENUM, subtree_ptr, subtree_len}
  - The last two elements would be a union.

alexisdal:
[...]

PR #52, opened 2017-11-06
Protocol Upgrade

samuelsadok:
Replace the user-oriented communication protocol with an improved machine oriented protocol.

The main differences are:
 - the device sends a JSON descriptor to announce it's parameters and functions to clients
 - CRC16 for UART
 - acknowledgements for parameter writes (and reads and function calls)

`test_communication.py` still provides the "p/v/c" commands but uses the new protocol internally. The tool now accepts a "--usb" and "--serial" parameter to select a specific device.

The old protocol is still operational on USB.


Firmware/MotorControl/crc.hpp
```diff
@@ -0,0 +1,66 @@
...
+//  source: https://users.ece.cmu.edu/~koopman/crc/index.html
+// Also known as CRC-16-DNP
+constexpr uint16_t CRC16_DEFAULT = 0x3d65;
+
+// Calculates an arbitrary CRC for one byte.
+// Adapted from https://barrgroup.com/Embedded-Systems/How-To/CRC-Calculation-C-Code
+template<typename T, unsigned POLYNOMIAL>
+static T calc_crc(T remainder, uint8_t value) {
+    constexpr T BIT_WIDTH = (CHAR_BIT * sizeof(T));
+    constexpr T TOPBIT = (1 << (BIT_WIDTH - 1));
```
kalvdans: Signed overflow if BIT_WIDTH is 32 or 64. Use `1ull << (BIT_WIDTH - 1)` or `(T)1 << (BIT_WIDTH - 1)` instead.

madcowswe:
Btw: This addresses issue #17.

Firmware/MotorControl/crc.hpp
samuelsadok: Good catch, fixed. For T=int32_t I don't see any issue though.

madcowswe: I think in the int32_t case, the signed overflow is technically undefined (though in practice it's bit-wise equivalent to unsigned overflow, which is defined).

https://stackoverflow.com/a/16188846/1128910

kalvdans (approved):

madcowswe:
Things to finish off:
- [x] Documentation on how to get started with the new ODrive Explore python
- [x] Clean out old communication Documentation (or move to depricated README page)
- [x] Note about how to switch to legacy protocol
- [x] Test that arduino library still works (Legacy on UART by default)

Issue #44, opened 2017-10-31
Fix printf deadlock


madcowswe:
`_write` in syscalls.c currently returns 0 when the transport is busy (both USB and UART). This is especially bad for USB, since the dffered IRQ pump is on the communication thread, the very thread that is likely to call printf (causing a deadlock in any retry loop).

Proposal: _write should block (wait for relevant semaphore) until the interface is ready. This is the expected behaviour for using printf. The caveat is that printf shouldn't be used in any realtime thread.

Optional: While we are at it, we can buffer the _write interface to make successive small writes more efficient. However then we need to be careful about how and when we flush the buffer, and by what thread and on what trigger conditions.

madcowswe:
Decision to not do buffering for now.
This issue is addressed by #57.

PR #57, opened 2017-11-08
Async USB Handling and printf blocking

Wetmelon:
Addresses #44

### DIFF
--- a/Firmware/MotorControl/commands.c
+++ b/Firmware/MotorControl/commands.c
@@ -22,7 +22,7 @@ SerialPrintf_t serial_printf_select = SERIAL_PRINTF_IS_NONE;
 static const GpioMode_t gpio_mode = GPIO_MODE_UART;     //GPIO 1,2 is UART Tx,Rx
 // static const GpioMode_t gpio_mode = GPIO_MODE_STEP_DIR; //GPIO 1,2 is M0 Step,Dir
 
-static uint8_t usb_buf[64];
+static uint8_t* usb_buf;
 static uint32_t usb_len;
 extern USBD_HandleTypeDef hUsbDeviceFS;
 
@@ -344,7 +344,8 @@ void cmd_parse_thread(void const * argument) {
                     }
                 }
             }
-            osStatus sem_stat = osSemaphoreWait(sem_usb_rx, 1);
+            // Check if there is USB processing to do.
+            osStatus sem_stat = osSemaphoreWait(sem_usb_rx, 0);
             if(sem_stat == osOK){
                 motor_parse_cmd(usb_buf, usb_len, SERIAL_PRINTF_IS_USB);
                 USBD_CDC_ReceivePacket(&hUsbDeviceFS); // Allow next packet
@@ -357,8 +358,8 @@ void cmd_parse_thread(void const * argument) {
 
 // Called from CDC_Receive_FS callback function, this allows motor_parse_cmd to access the
 // incoming USB data
-void set_cmd_buffer(const uint8_t *buf, uint32_t len) {
-    memcpy(usb_buf, buf, len);
+void set_cmd_buffer(uint8_t *buf, uint32_t len) {
+    usb_buf = buf;
     usb_len = len;
 }
 
--- a/Firmware/MotorControl/commands.c
+++ b/Firmware/MotorControl/commands.c
@@ -4,6 +4,7 @@
 #include <usart.h>
 #include <gpio.h>
 #include <freertos_vars.h>
+#include <usbd_cdc.h>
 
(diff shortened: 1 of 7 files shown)


Firmware/MotorControl/commands.c
```diff
@@ -335,21 +344,34 @@ void cmd_parse_thread(void const * argument) {
...
     }
-
     // If we get here, then this task is done
     vTaskDelete(osThreadGetId());
+}
+
+// Called from CDC_Receive_FS callback function, this allows motor_parse_cmd to access the
+// incoming USB data
+void set_cmd_buffer(const uint8_t *buf, uint32_t len) {
+    memcpy(usb_buf, buf, len);
```
madcowswe: We don't really need to do this copy, as we won't be recieving any new data to overwrite what's there until we have processed it. We can instead have the set_cmd_buffer just be setting a pointer to `UserRxBufferFS` instead of taking a copy.

Firmware/Src/syscalls.c
```diff
@@ -20,34 +21,40 @@
...
+            //Check length
+            if (len > UART_TX_BUFFER_SIZE)
+                return 0;
+            // memcpy data into uart_tx_buf
+            memcpy(uart_tx_buf, data, len);
+            // Start DMA background trasnfer
+            HAL_UART_Transmit_DMA(&huart4, uart_tx_buf, len);
+            
+            // Wait for the transmission to complete
+            osSemaphoreWait(sem_uart_dma, osWaitForever);
```
madcowswe: Wait on UART TX resource available before transmitting, instead of waiting for transmission to finish.

Firmware/Src/syscalls.c
```diff
@@ -20,34 +21,40 @@
...
+    //number of bytes written
+    int written = 0;
+    switch (serial_printf_select) {
+        case SERIAL_PRINTF_IS_USB: {
+            // transmit over CDC
+            uint8_t status = CDC_Transmit_FS((uint8_t*)data, len);
+            written = (status == USBD_OK) ? len : 0;
+
+            // Wait for transmission to complete
+            osSemaphoreWait(sem_usb_tx, osWaitForever);
```
madcowswe: Wait on USB TX resource available before transmitting, instead of waiting for transmission to finish.

Firmware/MotorControl/commands.c
```diff
@@ -335,21 +344,35 @@ void cmd_parse_thread(void const * argument) {
...
-            int USB_check_timeout = 1;
-            // Wait for signalling from USB interrupt (OTG_FS_IRQHandler)
-            osStatus semaphore_status = osSemaphoreWait(sem_usb_irq, USB_check_timeout);
-            if (semaphore_status == osOK) {
-                // We have a new incoming USB transmission: handle it
-                HAL_PCD_IRQHandler(&hpcd_USB_OTG_FS);
-                // Let the irq (OTG_FS_IRQHandler) fire again.
-                HAL_NVIC_EnableIRQ(OTG_FS_IRQn);
+            // Check if there is USB processing to do.
+            osStatus sem_stat = osSemaphoreWait(sem_usb_rx, 0);
```
madcowswe: This shouldn't be 0 wait, or else this thread will never sleep if there is nothing to do. You should make the timeout 1ms again, and put back the comment:

// When we reach here, we are out of immediate characters to fetch out of UART buffer
// Now we check if there is any USB processing to do: we wait for up to 1 ms,
// before going back to checking UART again.

Firmware/Src/syscalls.c
```diff
@@ -20,34 +21,39 @@
...
-  }
-
-  return written;
+int _write(int file, char* data, int len) {
+    //number of bytes written
+    int written = 0;
+    switch (serial_printf_select) {
+        case SERIAL_PRINTF_IS_USB: {
+            // Wait for the interface to be available
+            osStatus sem_stat = osSemaphoreWait(sem_usb_tx, osWaitForever);
```
madcowswe: I think it's safer to give it a 100ms timeout.

Firmware/Src/syscalls.c
```diff
@@ -20,34 +21,39 @@
...
+                written = (status == USBD_OK) ? len : 0;
+            }
+        } break;
+
+        case SERIAL_PRINTF_IS_UART: {
+            //Check length
+            if (len > UART_TX_BUFFER_SIZE)
+                return 0;
+            // Wait for the interface to be available
+            osStatus sem_stat = osSemaphoreWait(sem_uart_dma, osWaitForever);
```
madcowswe: I think it's safer to give it a 100ms timeout.

Issue #54, opened 2017-11-08
Unique Identification of ODrives


madcowswe:
Any system that uses multiple ODrives will need a way to repeatably tell them apart. We can use the 96bit unique identifier register that is available in the STM chips for this.

madcowswe:
Adressed by #59

PR #59, opened 2017-11-09
Add UUID access, echo cmd

Wetmelon:
Also makes the console a bit easier to read when handling a bunch of messages.

Addresses #54

### DIFF
--- a/Firmware/MotorControl/utils.h
+++ b/Firmware/MotorControl/utils.h
@@ -75,7 +75,7 @@
 // as per the magnitude invariant clarke transform
 // The magnitude of the alpha-beta vector may not be larger than sqrt(3)/2
 // Returns 0 on success, and -1 if the input was out of range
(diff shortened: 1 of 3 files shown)


Firmware/MotorControl/utils.h
```diff
@@ -13,11 +75,11 @@
 // as per the magnitude invariant clarke transform
 // The magnitude of the alpha-beta vector may not be larger than sqrt(3)/2
 // Returns 0 on success, and -1 if the input was out of range
-int SVM(float alpha, float beta, float* tA, float* tB, float* tC);
+int SVM(float alpha, float beta, float *tA, float *tB, float *tC);
```
madcowswe: why did you change this?

Wetmelon: I auto-formatted the document to clean up the defines and I guess it didn't like where the *'s were.

PR #62, opened 2017-11-14
Add CAN-bus commands support

fancer:
Hello Oscar,
In our project we needed the Odrive with 500Kb CAN-bus support. So for the test purpose I added a simple interface to handle ODrive commands coming over the CAN-bus. It supports position/velocity and current set commands, nothing else (see README for details). I think it should be enough for the most of the cases. Debug printfs still can be done via USB/UART. 
Should you have any questions, you can post them either here or address me via Skype: fancer.lancer


madcowswe:
Hi fancer,
Thanks so much for your contribution, implementing basic CAN support is very valuable.
I will do a more in depth review within a few days, but in the mean time, can you please do the following?
- [ ] Please target the madcowswe:devel branch
- [ ] I would suggest that you make your PR from a branch which isn't your master branch. The reason is that if you end up making other unrelated changes to your master branch , these would get sucked into the PR. So making a dedicated branch for the PR is recommended.
- [ ] Please resolve any new merge conflicts when you retarget to madcowswe:devel.

Again, thanks so much!

fancer:
Alright, I'll switch the target branch then. Alas I can't change the source branch without recreating the PR. The fork was created for just CAN interface development anyway, so I didn't need a separate branch in the first place. I merged all the changes with master yesterday, the new conflicts are related with last day commits. Ok, I'll fix them.

fancer:
Done.

Regarding the new alterations. I ported my changes in accordance with your USB-event-handler pattern IRQ->DisableIRQ->ReleaseSem->UnblockTask->HandleData->EnableIRQ. In future I'd suggest to move from such pattern either to IRQ->RecvData->PutToQueue->UnblockTask->HandleData, or IRQ-less pattern like RecvData->HandleData. It would let us to get free of IRQ-Enable/Disable calls, which in case of heavy traffic may lead to USB/CAN blocks Fifo overflow. And use just one structure for all incoming commands.

Additionally you added two new semaphores usage to synchronize access to the USB OTG. I don't know the whole reason of it, but according to my experience if one getting to use too many semaphores, it means something isn't right. I suppose we can use just one mutex to protect some critical section, so the write operation would be free of deadlock. I may be mistaken though...

-Sergey

Wetmelon:
@fancer Re the USB handling / semaphores / IRQ stuff - we had a long discussion about this while developing it.  Hop on our Discord and we'll be happy to explain our reasoning.  I'd also like to hear how you'd make it better!  I'm not exactly a USB expert, and it's my first time programming on a multithreaded system of any sort, so if you have suggestions that would be very helpful!

https://discord.gg/QKftWVc

I have a module I can use for injecting CAN messages and testing your PR, but it'll take me probably a week or so.  I assume it's been working well for you?

fancer:
@Wetmelon 
Yeh, the CAN interface is working well. I've used 8devices usb2can adapter together with Linux SocketCAN+canutils to test it out.

Issue #69, opened 2017-11-19
Current scale error compensation


madcowswe:
All ODrives v3.3 and earlier have an error on the board that causes a scale error in the current sensing. We should be able to compensate for it in firmware.

Now that the measurable scale is different for each board revision, we need to clip the `current_lim` to the measurable value. Before this was done with a comment telling the user to not set a value larger than 75A, but it's not that simple anymore.

madcowswe:
Implemented by #74

PR #74, opened 2017-11-23
Handle current scaling factor for HW versions <= 3.3

Wetmelon:
Addresses #69 - invisibly handles the current scaling, so users can still assume < 75A (if shunt resistor values are unchanged)

### DIFF
--- a/Firmware/MotorControl/low_level.c
+++ b/Firmware/MotorControl/low_level.c
@@ -421,7 +421,7 @@ static void DRV8301_setup(Motor_t* motor) {
 
         float margin = 0.95f;
         float max_input = margin * 0.3f * motor->shunt_conductance;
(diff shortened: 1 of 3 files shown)


Wetmelon:
I don't have a current clamp meter, so someone will have to verify this works correctly for ODrive 3.3 and 3.4.

Wetmelon:
After discussion with @madcowswe, I've modified the current_control struct to hold a "max_allowable_current" value which is calculated based on the shunt_conductance and the selected DRV amplifier gain.  Current is always clamped to the MIN of current_lim (selectable by the user) and max_allowable_current (calculated).

Shunt resistance (and therefore conductance) is now selected automatically by HW version.  It is 500ohms for 3.4 or newer, and 666 ohms for 3.3 or older.

Issue #38, opened 2017-10-10
[Feature Request] - Function to Set Encoder Register


youngiif:
I need a function to write to the encoder position register.

At least two threads on discourse discuss this desire:
https://discourse.odriverobotics.com/t/motor-rotor-encoder-state-read-only/269
https://discourse.odriverobotics.com/t/startup-against-gravity-with-encoder-with-index-pulse/251

I don't mind trying to create this feature and submit a pull request, I could just use a hand in getting started as I have no experience writing firmware for ARM chips.  I hopefully can trace/follow existing code enough to figure it out though.

If anyone can offer a bit of guidance, I am assuming I would need to create a function that calls DRV8301_writeSpi()?

madcowswe:
Thank you for volunteering to have a go at creating this.
So basically the 32 bit variable motor->encoder.encoder_state tracks the 16 bit count register in the STM32's timer pheriperal, located at motor->motor_timer.Instance->CNT.
The tracking is done [here](https://github.com/madcowswe/ODriveFirmware/blob/devel/MotorControl/low_level.c#L1107-L1108).

If you wish to update the encoder state, you must make sure to update the underlying register as well as the encoder_state.

madcowswe:
#75 implements this.
@youngiif can you please verify that this is able to do what you need?

PR #75, opened 2017-11-23
Create Set Encoder Count function

Wetmelon:
Addresses #38

### DIFF
--- a/Firmware/MotorControl/low_level.c
+++ b/Firmware/MotorControl/low_level.c
@@ -1380,8 +1380,9 @@ void motor_thread(void const * argument) {
     motor->thread_ready = false;
 }
 
-/** Function that sets the current encoder count to a desired 32-bit value. */
+// Function that sets the current encoder count to a desired 32-bit value.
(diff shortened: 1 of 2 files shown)


Firmware/MotorControl/low_level.c
```diff
@@ -1379,3 +1379,10 @@ void motor_thread(void const * argument) {
     }
     motor->thread_ready = false;
 }
+
+/** Function that sets the current encoder count to a desired 32-bit value. */
+void setEncoderCount(Motor_t* motor, uint32_t count){
+    motor->encoder.encoder_state = count;
+    motor->motor_timer->Instance->CNT = count;
```
madcowswe: There is a race condition hazard here: If this function is running on a low priority thread (say comms thread), it could in an unfortunate timing event get preempted right between these two lines, and have the motor thread's update_rotor run in-between. That would not be good.
We can make a short critical-section. We can check if the current OS api has an abstracted critical section. Otherwise, critical sections are usually implemented disable all interrupts (global interrupt mask), some critical code, enable them again.

PR #82, opened 2017-12-11
DFU over USB

samuelsadok:
Firmware: The command `enter_dfu_mode` makes the device enter the STM32 built-in USB DFU mode.
Python: `tools/dfu.py Firmware/build/ODriveFirmware.hex` puts the device into DFU mode and uploads the hex file.
Makefile: `make dfu` builds the hex file and uploads it to the device.

Known Issues:
  - The device can be bricked (e.g. on power loss). Unbricking requires an external programmer (STLink)
  - No signature checking
  - If multiple devices are connected, it's undefined which one will be upgraded


Wetmelon:
> If multiple devices are connected, it's undefined which one will be upgraded

I don't think that's acceptable.  We have several users with multiple boards hooked up to USB.  We have the ability to fetch an ODrive's GUID, we should use it (somehow)

samuelsadok:
Ok I wasn't aware that that would be a current issue. I assume you're referring to [this command](
https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/legacy_commands.c#L149)?

- [x] expose the STM UUID on an endpoint of the new protocol (`commands.cpp`)
- [x] check if the same UUID is also exposed by the STM's built-in DFU mode (probably it is).
- [x] allow specifying a UUID for device discovery (`core.py`) and add corresponding command line option to `dfu.py` and `make dfu`.

samuelsadok:
The serial number can now be specified like this:
```
make dfu SERIAL_NUMBER=385F324D3037
```
where `385F324D3037` is the serial number as shown in `lsusb`. For a given device this serial number is the same in both normal and DFU mode.

The serial numbers of all flashable devices can be listed by running `(lsusb -d 1209:0d32 -v; lsusb -d 0483:df11 -v) | grep iSerial`.

madcowswe:
@samuelsadok can you please add some notes in the readme about specifying the serial number for dfu?

madcowswe:
Two problems on windows:
1. after putting the ODrive in DFU mode, the STM32 BOOTLOADER won't show up in device mananger until you unplug/replug the usb cable
2. After doing above step, and the bootloader is available, the dfu script doesn't find it.

Also for convenience, the scan-for-device loop should include both bootloader and regular ODrive (presently it only looks for bootloader once).

samuelsadok:
* ~[ ] add check that `readline` is installed and add note to readme~ done on `sam_python_fixes`

samuelsadok:
ok ready for another review pass

PR #86, opened 2017-12-15
Feature/index pin

Wetmelon:
Implements the interrupt logic for using encoders with index lines.  Still need to write the calibration logic that finds the index pin on startup.


madcowswe:
Superseded by #106

PR #95, opened 2018-01-11
Added MacOS configuration for VSCode

kylebme:
The Linux and Windows configurations don't have the right file locations for the gcc-arm-embedded files. I added a configuration for Mac with the correct paths when the gcc arm toolchain was installed with Homebrew, as shown in the docs. The exact paths will change as it gets updated, but this gets people closer to the right paths.

This is my first PR, so I figured I would do something small :)


madcowswe:
Cool, thanks for contributing! ;D

PR #100, opened 2018-01-17
Fixed sending all commands to motor 0

fluture:


Wetmelon:
Can confirm this fixed my issue.

PR #113, opened 2018-02-19
updated readme to include mention of error codes

Capo01:
It may also be useful to add a short description of the common causes of each error code so that people can attempt to problem solve their own issues before asking for help.


Wetmelon:
Errors (ERROR_NO_ERROR) start at 0

PR #114, opened 2018-02-20
Feature/cpr check

Wetmelon:
Checks the encoders during calibration to try to give the user a head's up if they've got the wrong encoder CPR set.  Default threshold is 2% deviation from expected.

### DIFF
--- a/Firmware/MotorControl/low_level.c
+++ b/Firmware/MotorControl/low_level.c
@@ -129,6 +129,7 @@ Motor_t motors[] = {
             .encoder_offset = 0,
             .encoder_state = 0,
             .motor_dir = 1,   // 1 or -1
+            .encoder_calib_range = 0.02,
             .phase = 0.0f,    // [rad]
             .pll_pos = 0.0f,  // [rad]
             .pll_vel = 0.0f,  // [rad/s]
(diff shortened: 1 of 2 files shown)


Firmware/MotorControl/low_level.c
```diff
@@ -789,6 +792,14 @@ bool calib_enc_offset(Motor_t* motor, float voltage_magnitude) {
         }
         encvaluesum += (int16_t)motor->encoder.encoder_timer->Instance->CNT;
     }
+
+    float estEnc = scan_range / elec_rad_per_enc;
+    float adjustedCount = fabsf((int16_t)motor->encoder.encoder_timer->Instance->CNT)-init_enc_val;
```
madcowswe: I think the closing bracket of fabsf should be at the end, to include the subtraction?

madcowswe: Also I would maybe change the variables to be a bit more descriptive.
estEnc -> expected_enc
adjustedCount -> abs_delta_enc

Wetmelon:
> I think the closing bracket of fabsf should be at the end, to include the subtraction?

You mean final - initial works?  Man, I'm dumb lol.

> Also I would maybe change the variables to be a bit more descriptive.

Ok will do

PR #115, opened 2018-02-20
Random fixes

Wetmelon:
* Make M_PI a float literal (if utils.h is included)
* Allow gnu extensions
* Use the tag parser & fix the paths - no more uint32_t not a type error!

### DIFF
--- a/Firmware/.vscode/settings.json
+++ b/Firmware/.vscode/settings.json
@@ -3,7 +3,8 @@
     "files.exclude": {
         "build": true
     },
-    "C_Cpp.intelliSenseEngineFallback": "Disabled",
+    "C_Cpp.intelliSenseEngineFallback": "Enabled",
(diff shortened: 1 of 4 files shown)


Firmware/.vscode/settings.json
```diff
@@ -3,6 +3,8 @@
     "files.exclude": {
         "build": true
     },
+    "C_Cpp.intelliSenseEngineFallback": "Enabled",
+    "C_Cpp.intelliSenseEngine": "Tag Parser",
```
madcowswe, line 7: I'd like to get the Default engine working, the tag parser just gives me every single name ever, with no intelligence.
Can you see if you can make it work?
https://github.com/Microsoft/vscode-cpptools/blob/master/Documentation/LanguageServer/FAQ.md

PR #117, opened 2018-02-21
DRV Status reporting updates + microsecond delay function

qjones81:
Includes a couple of options for the status reporting.  The DRV8301 driver is not currently keeping the actual register value anywhere in memory but remaps to a struct.  So I added a value for each of the 4 regs so they could just be sent over as is when requested.  They are placed in the _DRV_SPI_8301_Vars_t_ struct that is currently cached in the Motor_t struct.  I also added caching for the DRV fault code that are in the DRV8301 driver as another option as well.  Finally included a drv_status.py file for demonstration.

Please take a look at the delay_us() code in utils.  I was running low on time so just did it with the Systick.  Timing looks accurate to me but probably needs to be evaluated later if there is a more reliable method.


madcowswe:
Looks great, can you please also add a note in the CHANGELOG.md about your feature?

PR #123, opened 2018-02-26
Prerelease fixes

Wetmelon:


Firmware/.vscode/tasks.json
```diff
@@ -6,7 +6,7 @@
         {
             "taskName": "build",
             "type": "shell",
-            "command": "make -j4",
+            "command": "make -j8",
```
madcowswe: Actually Tup will fire up an intelligent number of threads by itself, so we can drop the -j parameter alltogether.

PR #126, opened 2018-03-04
Fw v0.3.5 rc

madcowswe:
# Release candidate

### Added
* Reporting error if your encoder CPR is incorrect
* Ability to start anticogging calibration over USB protocol
* Reporting of DRV status/control registers and fault codes
* DRV status read script
* Microsecond delay function
* Travis-CI

### Changed
* Build system is now tup instead of make. Please check the [Readme](README.md#installing-prerequisites) for installation instructions.


Firmware/MotorControl/utils.c
```diff
@@ -183,3 +184,23 @@ uint32_t timeout_to_deadline(uint32_t timeout_ms) {
...
     return now_ms + timeout_ms;
 }
+
+// @brief: Returns number of microseconds since system startup
+uint32_t micros(void) {
+    uint32_t usTicks = HAL_RCC_GetSysClockFreq() / 1000000;
+    register uint32_t ms, cycle_cnt;
+    do {
+        ms = HAL_GetTick();
+        cycle_cnt = SysTick->VAL;
```
madcowswe: @qjones81 We don't use SysTick as timebase, we use TIM14. See the CubeMX project under SYS periperhal and/or `HAL_TIM_PeriodElapsedCallback` in main.c.
We should check if the autogenerated code somehow already emits a define or something like that for the timebase, otherwise we need to manually do it ourselves in main.h or something like that, such that this code can be easily maintained should we decide to switch the HAL timbase to another timer.

PR #128, opened 2018-03-05
Configurable Build

samuelsadok:
Let the user configure the build via the tup.config file.
With this change users should no longer need to edit source files.


madcowswe:
@samuelsadok Please add an entry to the Changelog.

samuelsadok:
done

tup.config.default
```diff
@@ -0,0 +1,6 @@
+CONFIG_DEBUG=y
+CONFIG_BOARD_VERSION=v3.3
+CONFIG_BOARD_48V=n
```
madcowswe: High voltage version might not always be 48V, I am already thinking of increasing it to 60ish V. So better call it HV.

Firmware/Tupfile.lua
```diff
@@ -1,7 +1,65 @@
...
+    FLAGS += "-DUART_PROTOCOL_LEGACY"
+elseif tup.getconfig("UART_PROTOCOL") == "none" then
+    FLAGS += "-DUART_PROTOCOL_NONE"
+else
+    error("unknown UART protocol "..tup.getconfig("UART_PROTOCOL"))
+end
+
+-- GPIO settings
+if tup.getconfig("STEP_DIR") == "y" then
+    if tup.getconfig("UART_PROTOCOL") != "none" then
```
madcowswe: Isn't the polarity of this check backwards?

Firmware/Tupfile.lua
```diff
@@ -1,7 +1,65 @@
 
 tup.include('build.lua')
 
-boarddir = 'Board/v3.3'
+-- Switch between board versions
+boardversion = tup.getconfig("BOARD_VERSION")
+if boardversion == "" then boardversion = "v3.4" end
```
madcowswe: It is dangerous to have a default like this. It hasn't happened yet, but it's concievable that the config of 1 board blows up another board. Better make sure people make it explicit.

tup.config.default
madcowswe: Make sure to also update readme.

PR #129, opened 2018-03-05
CubeMX cleanup

samuelsadok:
Regenerate CubeMX generated files with STM firmware 1.19.0 and add patchfiles that for changes that are otherwise overwritten by CubeMX. With this PR, rerunning CubeMX (with the same version) and afterwards applying the patches results in no changes.

 * [x] test if the above statement is true on Windows


madcowswe:
TODO:
- [x] Update .gitattributes to be in the local folder (it broke now with move v3.3 to v3).

PR #143, opened 2018-03-14
Safer malloc and use of core-coupled memory

samuelsadok:
* implement _sbrk_r syscall
* use core coupled memory for the FreeRTOS heap

~* [ ] test if CCM reduces jitter~ It does not, in fact the minimum runtime of the control loop is increased from ~2790 to ~2800 ticks. As per discussion we will use the CCM anyway to free up more space in the regular RAM


Firmware/Board/v3.3/Src/syscalls.c
```diff
@@ -19,6 +19,45 @@
...
+* Note that if this function returns NULL, malloc does not
+* consider this as an error and will return the pointer 0x8.
+*
+* You should still be careful with using malloc though,
+* as it does not guarantee thread safety.
+*
+* @return A pointer to the newly allocated block on success
+*         or -1 otherwise.
+*/
+intptr_t _sbrk_r(struct _reent *ignored __attribute__((unused)), size_t size) {
```
madcowswe: Why do we implement _sbrk_r instead of _sbrk?

samuelsadok: By default the call chain goes like this:
malloc => _malloc_r => _sbrk_r => _sbrk
In ODriveFirmware.elf there is no other place besides _sbrk_r that calls _sbrk directly, so currently overriding _sbrk_r has the same effect but keeps the option open to use the reent argument later.

Arguably we could also override _malloc_r such that it uses the FreeRTOS heap. The reasons why I didn't do that are:
 * I only noticed this option just now
 * If we go forward with CCM there'd me no more way to use the remaining space in main RAM

PR #153, opened 2018-03-28
Add Board & Firmware versions

samuelsadok:
### Added
 * `make write_otp` command to burn the board version onto the ODrive's one-time programmable memory. If you have an ODrive v3.4 or older, you should run this once for a better firmware update user experience in the future. Run the command without any options for more details. Once set, the board version is exposed through the `board_version_[...]` properties.
 * bake Git-derived firmware version into firmware binary. The firmware version is exposed through the `fw_version_[...]` properties.

### Changed
 * The DFU script now verifies the flash after writing


Firmware/MotorControl/commands.cpp
```diff
@@ -113,12 +114,40 @@ void enter_dfu_mode() {
     NVIC_SystemReset();
 }
 
+#if HW_VERSION_MAJOR == 3
+const uint8_t* otp_ptr =
+    *(uint8_t*)0x1fff7800 == 0xfe ? (uint8_t*)0x1fff7800 :
+    *(uint8_t*)0x1fff7800 != 0x00 ? NULL :
+        *(uint8_t*)0x1fff7810 == 0xfe ? (uint8_t*)0x1fff7810 : NULL;
```
madcowswe: What is going on here?

Firmware/MotorControl/commands.cpp
```diff
@@ -113,12 +114,40 @@ void enter_dfu_mode() {
...
+
+// Read hardware version from OTP if available, otherwise fall back
+// to software defined version.
+const uint8_t board_version_major = otp_ptr ? otp_ptr[3] : HW_VERSION_MAJOR;
+const uint8_t board_version_minor = otp_ptr ? otp_ptr[4] : HW_VERSION_MINOR;
+const uint8_t board_version_variant = otp_ptr ? otp_ptr[5] : (HW_VERSION_VOLTAGE == 24 ? 0 : 1);
```
madcowswe: Comparing with exactly 24 is not very stable, the nominal voltage may be different in the future.
Why not just expose the voltage instead of a variant-number?

madcowswe (commented):
After you have addressed this, let's do a OTP setting cycle together.

Issue #66, opened 2017-11-19
Absolute encoder over SPI/SSI


madcowswe:
We want asynchronous: Simple fire and check complete flag and read from register later if the encoder value is a single frame, DMA if not. 16bit single frame should be fine for the vast majority of SPI/SSI encoders.

We also want it to be triggered at a very defined time. If possible we want either the TIM1 (or 8) to hardware trigger the transfer, or we need to manually trigger it at the very beginning of the adc_pwm_cb code, to reduce jitter

Wetmelon:
Per discourse discussion [here](https://discourse.odriverobotics.com/t/suggestions-for-a-new-encoder/319/4) the AMT10x series are being tested.  CUI also makes a very similar AMT20x series, which is essentially a 10-bit AMT10x with an additional 12-bit absolute encoder accessible over SPI.  I'm going to grab one of these for testing purposes.

There is also the [CTR Electronics SRX Magnetic Encoder](http://www.ctr-electronics.com/downloads/pdf/Magnetic%20Encoder%20User's%20Guide.pdf), which is what's used in the [VersaPlanetary Integrated Encoder](https://www.vexrobotics.com/vexpro/motors-electronics/encoders/217-5046.html).  This is a "1024CPR" (unsure if it's actually 1024 ppr or cpr) incremental encoder with an additional 244Hz PWM signal for absolute positioning.  Because of the simple integration with the VersaPlanetary gearbox, this might also be a good option.

PR #163, opened 2018-04-28
Support for AMT203 absolute encoder

Wetmelon:
Provides full SPI support for the AMT203 encoder, and defines an interface for all absolute encoders in general.

If the ODrive is configured to use the AMT203 encoder (encoder.config.encoder_type = ENCODER_TYPE_ABSOLUTE_AMT203), it will construct the object and initialize communications.  This includes setting the chip select pin high soas not to interfere with the DRV8301 comms.  What follows is dependent on configuration settings, and is described in the following code snippet:

```C++
    if (abs_enc_ != nullptr) {
        abs_enc_->init();
        if (config_.amt203_config.use_absolute) {
            set_circular_count(abs_enc_->readPosition());
            if(config_.pre_calibrated) {
                offset_ = config_.offset;
                is_ready_ = true;
            }
        }
    }
```
This initialization is done per-axis, in sequence.  After setting the offset, the quadrature interface is used exclusively for positioning.

SPI semaphores were added for future async SPI support.

Addresses #66

### DIFF
--- a/Firmware/Encoders/amt203.cpp
+++ b/Firmware/Encoders/amt203.cpp
@@ -1,0 +1,111 @@
+#include "amt203.hpp"
+#include "utils.h"
+
+SPI_HandleTypeDef spiSettings_ = {
+    .Instance = SPI3,
+    .Init = {
+        .Mode = SPI_MODE_MASTER,
+        .Direction = SPI_DIRECTION_2LINES,
+        .DataSize = SPI_DATASIZE_8BIT,
+        .CLKPolarity = SPI_POLARITY_LOW,
+        .CLKPhase = SPI_PHASE_2EDGE,
+        .NSS = SPI_NSS_SOFT,
+        .BaudRatePrescaler = SPI_BAUDRATEPRESCALER_16,
+        .FirstBit = SPI_FIRSTBIT_MSB,
+        .TIMode = SPI_TIMODE_DISABLE,
+        .CRCCalculation = SPI_CRCCALCULATION_DISABLE,
+        .CRCPolynomial = 10}};
+
+// See http://www.cui.com/product/resource/AMT203-v.pdf for protocol information
+AMT203::AMT203(SPI_HandleTypeDef* spiHandle, GPIO_TypeDef* chipSelectHandle, uint16_t chipSelectPin)
+    : spiHandle(spiHandle),
+      chipSelectHandle(chipSelectHandle),
+      chipSelectPin(chipSelectPin),
+      position(0) {
+    spiSettings_.Instance = spiHandle->Instance;
+}
+
+bool AMT203::init() {
+    // Initialize the chip select pin so it behaves how the AMT203 expects
+    HAL_GPIO_DeInit(chipSelectHandle, chipSelectPin);
+    GPIO_InitTypeDef GPIO_InitStruct;
+    GPIO_InitStruct.Pin = chipSelectPin;
+    GPIO_InitStruct.Mode = GPIO_MODE_OUTPUT_PP;
+    GPIO_InitStruct.Pull = GPIO_PULLUP;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(chipSelectHandle, &GPIO_InitStruct);
+
+    // Set chipSelect high
+    HAL_GPIO_WritePin(chipSelectHandle, chipSelectPin, GPIO_PIN_SET);
+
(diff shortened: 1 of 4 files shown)


Firmware/Encoders/amt203.cpp
```diff
@@ -0,0 +1,111 @@
...
+    GPIO_InitStruct.Pull = GPIO_PULLUP;
+    GPIO_InitStruct.Speed = GPIO_SPEED_FREQ_LOW;
+    HAL_GPIO_Init(chipSelectHandle, &GPIO_InitStruct);
+
+    // Set chipSelect high
+    HAL_GPIO_WritePin(chipSelectHandle, chipSelectPin, GPIO_PIN_SET);
+
+    // Reinitialize the SPI bus to the required settings,
+    // but use the same Instance that was passed to the object
+    HAL_SPI_DeInit(spiHandle);
```
samuelsadok, line 43: I think the pattern
```
HAL_SPI_Init(spiHandle);
[do SPI stuff]
HAL_SPI_DeInit(spiHandle);
```
would be more intuitive

Firmware/Encoders/amt203.cpp
```diff
@@ -0,0 +1,111 @@
...
+    *spiHandle = spiSettings_;
+    HAL_SPI_Init(spiHandle);
+
+    //! WARNING! Potential for thread-lockup!
+    while (sendNop() != AMT203_NOP_A5) {
```
samuelsadok, line 48: given that this wait loop occurs multiple times, you can pull it out into a function and conveniently add a timeout check. Something like:
```
bool wait_loop(uint32_t timeout_ms) {
    deadline = deadline_to_timeout(timeout_ms);
    while (sendNop() != AMT203_NOP_A5) {
        if (!is_in_the_future(deadline))
            return false;
        // maybe also add an osDelay(1) to free up CPU time? (don't know if 1ms is too much)
    }
    return true;
}
```
The `is_in_the_future()` function you can copy from [this feature branch](https://github.com/madcowswe/ODrive/blob/sam_can/Firmware/MotorControl/utils.c#L182-L187)

Firmware/build.lua
```diff
@@ -79,8 +79,8 @@ function GCCToolchain(prefix, builddir, compiler_flags, linker_flags)
         }
     end
     return {
-        compile_c = function(src, flags, includes, outputs) gcc_generic_compiler(prefix..'gcc -std=c99', compiler_flags, true, src, flags, includes, outputs) end,
-        compile_cpp = function(src, flags, includes, outputs) gcc_generic_compiler(prefix..'g++ -std=c++14', compiler_flags, true, src, flags, includes, outputs) end,
+        compile_c = function(src, flags, includes, outputs) gcc_generic_compiler(prefix..'gcc -std=gnu11', compiler_flags, true, src, flags, includes, outputs) end,
+        compile_cpp = function(src, flags, includes, outputs) gcc_generic_compiler(prefix..'g++ -std=gnu++17 -Wno-register', compiler_flags, true, src, flags, includes, outputs) end,
```
samuelsadok: IIRC this was just for using std::variants, but you don't use them now. So I guess we can stay at C++14?

Firmware/MotorControl/main.cpp
```diff
@@ -108,6 +108,19 @@ int odrive_main(void) {
     for (size_t i = 0; i < AXIS_COUNT; ++i) {
         Encoder *encoder = new Encoder(hw_configs[i].encoder_config,
                                        encoder_configs[i]);
+
+        switch (encoder->config_.encoder_type) {
```
samuelsadok: I guess this can be moved into the `Encoder` constructor

Firmware/Encoders/amt203.cpp
```diff
@@ -0,0 +1,111 @@
...
+ *!  the position values will not be calculated off the latest zero position. */
+bool AMT203::setZeroPoint() {
+    uint8_t response = sendCommand(AMT203_SET_ZERO_POINT_CMD);
+
+    //! WARNING! Potential for thread-lockup!
+    do {
+        response = sendNop();
+    } while (response == AMT203_NOP_A5);
+
+    if (response == AMT203_SET_ZERO_POINT_SUCCESS)
```
samuelsadok, line 94: `return response == AMT203_SET_ZERO_POINT_SUCCESS;`

samuelsadok (commented):
Just a few minor comments, otherwise lgtm.
Maybe we should also allow the user to recalibrate the absolute position at will (using a remote function call) but that would only be required for long running systems.

PR #164, opened 2018-05-13
DFU Improvements

samuelsadok:
* Automatically download firmware from GitHub releases if no file is provided
 * Retain configuration during firmware updates
 * Add commands `odrivetool backup-config` and `odrivetool restore-config`

### DIFF
--- a/tools/odrive/dfu.py
+++ b/tools/odrive/dfu.py
@@ -9,8 +9,9 @@ import time
 import threading
 import platform
 import struct
-import array
-import fractions
+import requests
+import re
+import io
 import usb.core
 import odrive.discovery
 from odrive.utils import Event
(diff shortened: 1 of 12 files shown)


tools/odrive/dfu.py
```diff
@@ -161,42 +74,136 @@ def get_first_mismatch_index(array1, array2):
...
+    data = dfudev.read_sector(otp_lock_sector)
     print(' '.join('{:02X}'.format(x) for x in data))
 
+class Firmware():
+    def __init__(self):
+        self.fw_version = (0, 0, 0, True)
+        self.hw_version = (0, 0, 0)
+
+    @staticmethod
+    def is_newer(a, b):
```
madcowswe, line 99: Tuple comparison is built in

tools/odrive/dfu.py
```diff
@@ -161,42 +74,136 @@ def get_first_mismatch_index(array1, array2):
...
-    status = dfudev.wait_while_state(DfuState.DFU_MANIFEST_SYNC)
-    if status[1] != DfuState.DFU_MANIFEST:
-        raise RuntimeError("An error occured. Device Status: {}".format(status[1]))
-
-
-def dump_otp():
+def dump_otp(dfudev):
     """
     Dumps the contents of the one-time-programmable
     memory. The OTP will be used in future versions of
```
madcowswe: Not future anymore

tools/odrive/dfu.py
samuelsadok: addressed

Issue #131, opened 2018-03-06
Better compatibility/user experience for libusb vs com port.


madcowswe:
Having to use Zadig on Windows is a bit annoying, and we can't even switch to native on Mac.
We would like to simultaneously (without recompiling) support both stream based and packet based native protocol.

We can either expose two endpoints with different capability manifest (one says they are a CDC, other is "custom"), or we make it easy to switch from one to the other.

madcowswe:
Dual interfaces implemented in #166. Zadig still needed, we can fix this later.

Issue #97, opened 2018-01-12
USB Protocol doesn't work on Mac


kylebme:
I'm opening a separate issue for this protocol. This is for the USB_PROTOCOL_NATIVE protocol.

This protocol doesn't work with Mac. The issue seems to be because the operating system matches the device with a built-in driver. PyUSB/libusb can't connect to it because it is already connected to a driver. MacOS doesn't seem to let libusb disconnect the kernel driver from it, the way it can in Linux.

I've tried unloading both the official FTDI driver (/Library/Extensions/FTDIUSBSerialDriver.kext) and the built-in Mac FTDI driver (/System/Library/Extensions/AppleUSBFTDI.kext). This doesn't fix it. I did notice a kext named com.apple.driver.usb.cdc, but it wouldn't let me unload it ("in use or retained").

Unloading common USB drivers seems like a poor fix for this. The real fix is to make, sign, and distribute a "codeless kext", which doesn't do anything at all, but that matches the device by id, name, etc. That prevents the system driver from connecting.

Some of these methods are mentioned here: https://github.com/libusb/libusb/wiki/FAQ#How_can_I_run_libusb_applications_under_Mac_OS_X_if_there_is_already_a_kernel_extension_installed_for_the_device

Also, I don't know if these solutions will work at all. There could be some other issue that will prevent this protocol from working with Mac.

kylebme:
Here's some info from the Discord about the way this protocol was devised, from Oskar:

> ok, so here's how it goes:
The odrive gives the PC a USB descriptor that says "I'm a CDC device, here are my standard capabilities"
the PC is like cool, we know what to do here, lets use the OS provided CDC driver: the ODrive shows up as a serial port
as we send data as a serial port, the PC packetises it up and sends it over the CDC spec to the ODrive
STM doesn't actually assemble it back into a character stream for you, they actually only help making the descriptor. The callbacks just give you a raw data buffer of what was sent in that specific packet.
so we were like, huh whatever, let's just send data with packets then, we'll get more control anyway
so we (on Windows and Linux) tell the PC: No, we actually don't want the OS's CDC driver, we'll do this ourselves with libusb/pyusb
so basically we use STM's CDC setup to get everuthing up and running, then we go "jk, let's do it raw packets"
There are two cleaner alternatives of course:

[...]

PR #166, opened 2018-05-14
USB composite device

samuelsadok:
* The device now appears as a composite device on USB. One subdevice is still a CDC device (virtual COM port), the other subdevice is a vendor specific class. This should resolve several issues that were caused by conflicting kernel drivers or OS services.
* Add WinUSB descriptors. This will tell Windows >= 8 to automatically load winusb.sys for the ODrive (only for the vendor specific subdevice). This makes it possible to use the ODrive from userspace via WinUSB with zero configuration. The Python tool currently still uses libusb so Zadig is still required.
* Add a configuration to enable the ASCII protocol on USB at runtime. This will only enable the ASCII protocol on the USB CDC subdevice, not the vendor specific subdevice so the python tools will still be able to talk to the ODrive.

addresses #131 and #97


kylebme:
I will test this on MacOS this weekend.

PR #170, opened 2018-06-04
AS5047P

aksatshah:


Firmware/Drivers/AS5047P/as5047p.c
```diff
@@ -0,0 +1,43 @@
...
+	return data;
+
+}
+
+
+uint16_t AS5047P_readSpi(AS5047D_Handle handle, uint16_t command){
+	HAL_GPIO_WritePin(handle->nCSgpioHandle, handle->nCSgpioNumber, GPIO_PIN_RESET);
+
+	uint16_t buff = 0;
+
```
madcowswe, line 36: Are you observing the setup and hold times here? You may want a delay between the GPIO operation and the SPI operation, (and after SPI operation too)

Firmware/Src/freertos.c
```diff
@@ -153,6 +164,13 @@ void StartDefaultTask(void const * argument)
   thread_motor_0 = osThreadCreate(osThread(task_motor_0), &motors[0]);
   thread_motor_1 = osThreadCreate(osThread(task_motor_1), &motors[1]);
 
+  osDelay(200);
+
+  update_init_cnt_value(&motors[0]);
+  update_init_cnt_value(&motors[0]);
```
madcowswe, line 170: Typo, you probably mean motors[1] on the 2nd one.

madcowswe (commented):
Looks like the right approach in general, some comments on the implementation.

samuelsadok:
Another implementation for SPI encoder support (initial commit: 4b29c5154893f74a4ea4308ac6b0aaa02e006dbd) was now merged to `devel`.

PR #174, opened 2018-06-18
Fix odrive pip install

madcowswe:
remove install-time PySerial dependency and don't fail if installed as non-root


madcowswe:
Fixes released on pip version `0.4.0.post2`

Shreeyak:
I tried installing odrivetool via pip 1 day ago and faced the pywin32 dependency on an Ubuntu 16.04 machine.

PR #177, opened 2018-06-20
Integrated some of AS5047P as a new encoder mode and got it to compile.

maxbeeman:


madcowswe:
Please let us know on discord if you need any assistance or feedback. We'd love to have support for the AS5047P.

PR #178, opened 2018-06-20
Fix timeout error code for macOS

Nate711:
In macOS, libusb reports timeouts with error code 60, which is different from other platforms.
This fixes an issue where the ODrive would immediately disconnect after connecting to odrivetool. (Written by Sam)


madcowswe:
hotfix released on pip: odrive-0.4.0.post4

PR #189, opened 2018-06-27
Trapezoidal

IWenjunSun:
Trapzoidal


Wetmelon:
I don't get it, why's there so much code to generate a few slopes?

And why's the code unreadable?  It needs a once-over to make it self-commenting at the very least.

madcowswe:
Sorry, I cannot accept code that is this bloated, unstructured or unmaintainable. I know perhaps the intention wasn't to make it mergable to the ODrive trunk, and I think I misunderstood.
Thanks for making the PR when I asked anyway.

PR #191, opened 2018-06-28
AS5047P Absolute Encoder + Parallel PD Gain Function

DJGCrusader:
This integrates Aksat's work into getting the AS5047P Absolute Encoder working in SPI. Uses GPIO 3 and 4 for Slave Select of Encoder 1 and 2. I did this by adding an encoder mode in addition to Incremental and Hall. 

I also added a controller function to modify the PD gains in parallel form, then convert them to the presently implemented form. 

Some random default configuration settings, such as pll bandwidth and other things may have been changed to the setting I am using for my setup, but these may be disregarded.


samuelsadok:
Another implementation of SPI encoder support has been merged now.
Regarding the simultaneous setup of `pos_gain` and `vel_gain` is there a use case that requires setting these atomically? Because I think the direction we should take is to let user convenience functions live in the UI (odrivetool) rather than on the device itself.

DJGCrusader:
> Another implementation of SPI encoder support has been merged now.

Good, that's fine

> Regarding the simultaneous setup of `pos_gain` and `vel_gain` is there a use case that requires setting these atomically? Because I think the direction we should take is to let user convenience functions live in the UI (odrivetool) rather than on the device itself.

Legged robotics, where each joint you want to specify a stiffness and damping, and you want to change them on the fly. The nested loop architecture is more conducive to machine axis motion. Rather than have a CNC machine axis architecture and hack a way on top of it to get a specified impedance for my legged robot, having this option built-in would make debugging and control design easier.

PR #193, opened 2018-07-01
Update odrive/docs/encoders.md to fix typo

Nate711:
The documentation mistakenly says to set <axis>.encoder.requested_state, when it should be <axis>.requested_state.


madcowswe:
Thanks!

PR #196, opened 2018-07-17
Docs: Add note for error I received and resolution

metanoic:
Adding tip for anyone running into the issue I ran into of openocd not being able to find its scripts directory/.cfg files.


madcowswe:
Thanks

PR #199, opened 2018-07-23
Python 2.7 compatibility

neomanic:
In preparation for a ROS driver for the ODrive, I've added Python 2.7 compatibility.

Tested on macOS and Ubuntu 16.04 via direct use and pip -e installs. I haven't tested building the distribution files.

Notable changes:
- Threading API constructor can't take the daemon parameter, so all thread creation had to be expanded out.
- TimeoutError isn't defined, but it makes for more readable code, so I defined it as an OSError subclass.
- ModuleNotFoundError is replaced by the older ImportError.
- Print function imported from future

### DIFF
--- a/tools/odrive/version.py
+++ b/tools/odrive/version.py
@@ -36,7 +36,7 @@ def get_version_from_git():
         print(ex)
         return "[unknown version]", 0, 0, 0, 1
 
-def get_version_str(git_only=False, is_post_release=False, bump_rev=False):
+def get_version_str(git_only=False, is_post_release=False, bump_rev=False, release_override=False):
     """
     Returns the versions of the tools
     If git_only is true, the version.txt file is ignored even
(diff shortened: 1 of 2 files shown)


neomanic:
Just noticed a few issues after generating the distribution tar.gz and installing from that. 

`odrivetool` has python3 set in the hash-bang header. I've been running it with `python2 odrivetool ...`. Not sure of the best option here, but anyone wanting use odrivetool completely under 2.7 will need to tweak this.

I also haven't yet used the DFU functionality, but that has a print function that needs updating to import from future.

`tests.py` uses type annotations, which aren't backwards-compatible, obviously. I'm thinking just to leave this alone, as I'm not too concerned with actually running the tests.

madcowswe:
Awesome thanks so much! It may take quite a while until we have enough time to get around to this PR, but py2 and ROS support is something we definitely want.

Note to self (and others who see this), @neomanic's ROS driver is here: https://github.com/neomanic/odrive_ros

neomanic:
I needed to have my own master branch on my fork so I can make some changes without requiring all my py2.7 compat work along for the ride, which is now on it's own branch. I'll reopen the pull request off the new branch, just ping me when you have the time to review.

PR #201, opened 2018-07-25
Update odrivetool

Capo01:
Note sure if this is intentional but currently liveplotter is listed as 'Upgrade the ODrive's Firmware' in the odrivetool help file. That doesn't seem right to me.


madcowswe:
Thanks!

PR #204, opened 2018-07-27
Expose crc as endpoint 1

metanoic:
This PR exposes the json_crc value as endpoint 1.  Endpoint 1 gets treated extra special in that it doesn't matter what the "trailer" or signature of the packet is.  e.g. for Endpoint 0 it expects the current protocol version, but to do so for endpoint 1 would require a change to fibre and would not be backwards compatible with prior firmware since it will be sending the crc instead of protocol version.

It also adds some notes to the area of code that sets up the endpoint members; instructing future developers to not re-arrange a group of members so that future revisions of the schema don't change endpoint numbers for members that are useful for determining board and firmware versions.

### DIFF
--- a/Firmware/communication/communication.cpp
+++ b/Firmware/communication/communication.cpp
@@ -8,6 +8,8 @@
 #include "interface_can.hpp"
 #include "interface_i2c.h"
 
+#include "fibre/protocol.hpp"
+
 #include "odrive_main.h"
 #include "freertos_vars.h"
 #include "utils.h"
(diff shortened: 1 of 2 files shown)


samuelsadok:
I believe the use case that this PR caters for is now addressed by #353.

As far as I can see, host software falls into one of these categories:
 - Works with any interface version. In this case it would need to download the complete JSON file.
 - Works with only one well known interface version. In this case it would know the JSON CRC beforehand and an ODrive with a mismatched interface version would simply not respond.
 - Works with a selection of well known interface versions. In this case the host would need a way to retrieve the interface version (which may be an opaque ID) which then implies the CRC. #353 addresses this by returning an opaque interface version ID when endpoint 0 is read at offset 0xffffffff.

metanoic:
Closing since this is now addressed in a cogent manner.  Thanks @samuelsadok, and it's nice to see you back!

PR #210, opened 2018-08-12
Trapezoidal Trajectory Generator

Wetmelon:
Implements a version of the Acceleration Limited trajectory algorithm described in https://hal.archives-ouvertes.fr/hal-01575057/document

Supports online kinematic boundary changes, as well as handling short moves and "overshoots".

Use:
1. Create Trapezoidal Trajectory object
2. Call planTrapezoidal once per move
3. Call evalTrapTraj with the current time 0 < t < Ta+Tv+Td

evalTrapTraj returns a struct with the position, velocity, and acceleration values defined by the precomputed trajectory.


Wetmelon:
This is ready for merge

Firmware/MotorControl/trapTraj.cpp
```diff
@@ -0,0 +1,112 @@
...
+            Ta = std::max(0.0f, (Vr - Vi) / Ar);
+            Tv = 0;
+            Td = std::max(0.0f, -Vr / Dr);
+        } else {
+            Tv = (dX - dXmin) / Vr;
+        }
+    }
+
+    // Overshoot move handling
+    if(std::abs(Xf - Xi) <= ((Vi * Vi) / (2.0f * Dmax))) {
```
madcowswe: This condition is identical to `std::abs(dX) <= dx_stop` on line 31. Consider joining this block with that one.

Firmware/MotorControl/trapTraj.cpp
```diff
@@ -0,0 +1,112 @@
...
+    } else if (t < Ta_ + Tv_) {  // Coasting
+        trajStep.Y = yAccel_ + (Vr_ * (t - Ta_));
+        trajStep.Yd = Vr_;
+        trajStep.Ydd = 0;
+    } else if (t < Ta_ + Tv_ + Td_) {  // Deceleration
+        float Tdc = t - Tav_;
+        trajStep.Y = yAccel_ + (Vr_ * (t - Ta_)) + Dr_ * (Tdc * Tdc) / 2.0f;
+        trajStep.Yd = Vr_ + Dr_ * Tdc;
+        trajStep.Ydd = Dr_;
+    }
```
madcowswe: Missing final condition

Firmware/MotorControl/trapTraj.cpp
```diff
@@ -0,0 +1,112 @@
...
+            Tv = (dX - dXmin) / Vr;
+        }
+    }
+
+    // Overshoot move handling
+    if(std::abs(Xf - Xi) <= ((Vi * Vi) / (2.0f * Dmax))) {
+        overshoot_move_ = true;
+        TrapTrajStep_t trajStep = evalTrapTraj(Ta+Tv+Td);
+        overshoot_time_ = Ta + Tv + Td;
+        return planTrapezoidal(Xf, trajStep.Y, trajStep.Yd, Vmax, Amax, Dmax); // Recalculate the move's second half
```
madcowswe: Neater recursive style is to do the `Ta += overshoot_time_;` stuff up here after evaluating the lower trajectory "bare", instead of pushing down overshoot_move_ and overshoot_time_

Firmware/MotorControl/trapTraj.cpp
Wetmelon: Good catch.  Moved.

Firmware/MotorControl/trapTraj.cpp
Wetmelon: Added final condition.

Firmware/MotorControl/trapTraj.cpp
Wetmelon: Not sure what you're referring to here.

Wetmelon:
An overshoot move occurs when  `abs(dX) <= dX_stop` where dX is the total move distance (start to finish) and dX_stop is the minimum stopping distance at maximum deceleration, given a non-zero starting velocity.

To force an overshoot move, start a long distance, relatively high velocity move with high accelerations.  When in the coast phase, reduce the `config.decel_limit`, and then call `move_to_pos` again with a stopping point that is too close for the motor to stop.

`abs(Xf-Xi) <= Vi*Vi / (2*Dmax)`

High starting velocity, short distance move, with low `decel_limit`.

madcowswe:
Let's merge #252 first, then we can update this to include the last pieces that were added after I started my review.

PR #216, opened 2018-08-18
Update getting-started.md

gpenzo:
On ubuntu mate the odrivetool needs root privileges else the odrive will not be found.


madcowswe:
I think that if you have set up udev rules this shouldn't be required? Do you have udev rules in place?

gpenzo:
I followed your guide.
My udev rules 
cat /etc/udev/rules.d/50-odrive.rules
SUBSYSTEM=="usb", ATTR{idVendor}=="1209", ATTR{idProduct}=="0d[0-9][0-9]", MODE="0666"
My firmware is based on a clone of saturday night, master branch,

Maybe you correct that this should work. Does not work here and for me it is no problem. 
Just made this change too inform others when it does not work. Maybe a warning with the risk of running stuff in root can make it clear..

madcowswe:
Superseeded by #241

Issue #219, opened 2018-08-20
Unable to use closed loop control after index only search on startup


Capo01:
I am unable to have the motor enter closed loop control after running only an index search. V3.2 board with AMT102 encoder running devel. All other control appears normal.

All settings erased before I entered the following commands into odrivetool as per the instructions:

```
odrv0.axis0.requested_state = AXIS_STATE_FULL_CALIBRATION_SEQUENCE
odrv0.axis0.encoder.config.use_index = True
odrv0.axis0.requested_state = AXIS_STATE_ENCODER_INDEX_SEARCH
odrv0.axis0.requested_state = AXIS_STATE_ENCODER_OFFSET_CALIBRATION
odrv0.axis0.motor.config.pre_calibrated = True
odrv0.axis0.config.startup_encoder_index_search = True
odrv0.axis0.config.startup_closed_loop_control = True
odrv0.save_configuration()
odrv0.reboot()
```

On reboot the motor does its index search but does not enter closed loop control. No error codes and control is in the idle state.
```
In [202]: hex(odrv0.axis0.error)
Out[202]: '0x1'
In [203]: hex(odrv0.axis0.encoder.error)
Out[203]: '0x0'
In [204]: hex(odrv0.axis0.motor.error)
Out[204]: '0x0'
In [205]: odrv0.axis0.motor.is_calibrated
Out[205]: True
```

Adding the following commands resolves the issue
```
odrv0.axis0.config.startup_encoder_offset_calibration = True
odrv0.save_configuration()
odrv0.reboot()
```

Now the motor first does its index search and then does its offset calibration. Only now is closed loop control entered. 

This does not seem to be correct to me as the motor should be usable without the need for the offset calibration once the index signal has been setup. If it is simply an error in the order that I am entering commands then perhaps the instructions may need to be changed.

Wetmelon:
I would try:

```
odrv0.axis0.encoder.config.use_index = True
odrv0.axis0.requested_state = AXIS_STATE_FULL_CALIBRATION_SEQUENCE
odrv0.axis0.motor.config.pre_calibrated = True
odrv0.axis0.config.startup_encoder_index_search = True
odrv0.axis0.config.startup_closed_loop_control = True
odrv0.save_configuration()
odrv0.reboot()
```

Capo01:
I erased the board and entered the commands in the order you suggested. The problem still persists. On reboot it moves to the index location but then fails to enter closed loop control.

[...]

PR #225, opened 2018-08-25
reordered index_search and offset_calibration in `AXIS_STATE_FULL_CAL…

Capo01:
…IBRATION_SEQUENCE`

As per issue #219, the index search position was not being recorded for some reason when the index search was conducted before the offset search. By reordering the commands the index position is now recorded when AXIS_STATE_FULL_CALIBRATION_SEQUENCE is called.


madcowswe:
The way the logic is set up, the index search should happen first, then the offset calibration. Your issue must be caused by some bug. Let's investigate.
Reordering isn't the correct solution, but thanks anyway.

PR #235, opened 2018-09-06
Added details about the PID control used by odrive

Capo01:
There was a brief discussion on discord about the PID control style used by odrive and Crinq made the comments used here. I thought they could be of interest for advanced users.

Just a thought, feel free to disregard if you think it does't fit with the rest of the docs.


metanoic:
~~Qrinq~~ Crinq

PR #236, opened 2018-09-07
Python 2.7 compatibility

neomanic:
In preparation for a ROS driver for the ODrive, I've added Python 2.7 compatibility. This updates an earlier pull request that I closed because I've moved this to it's own branch.

Tested originally on macOS and Ubuntu 16.04 via direct use and pip -e installs, but not recently since I just sent off my working ODrives this morning and need to check again. I haven't tested building the distribution files.

Notable changes:

- Threading API constructor can't take the daemon parameter, so all thread creation had to be expanded out.
- TimeoutError isn't defined, but it makes for more readable code, so I defined it as an OSError subclass.
- ModuleNotFoundError is replaced by the older ImportError.
- Print function imported from future

### Notes from previous pull request of things that still need to be done.

Just noticed a few issues after generating the distribution tar.gz and installing from that.

odrivetool has python3 set in the hash-bang header. I've been running it with python2 odrivetool .... Not sure of the best option here, but anyone wanting use odrivetool completely under 2.7 will need to tweak this.

I also haven't yet used the DFU functionality, but that has a print function that needs updating to import from future.

tests.py uses type annotations, which aren't backwards-compatible, obviously. I'm thinking just to leave this alone, as I'm not too concerned with actually running the tests.


madcowswe:
Thank you so much for this!

neomanic:
Thanks mate! Just one question before this is done and dusted; I neglected to do anything with tests (since they have type annotations as per comment above), but that could mean that going forward it'll require manual testing (from me and other ROS users) to ensure we retain Py2.7 compatibility. I imagine everyone else other than ROS users will be on Py3.5+.

So while I don't have the headspace to add tests right away, when it does come time would you prefer I:
1. create a separate tests_py27.py file, or
2. change the tests to use comment-based annotation as per PEP484.

I'd much prefer (2) since then it'll be a single file to support both platforms, but if you prefer the cleanliness of inline annotations I understand.

madcowswe:
I agree that it's important to test on all supported platforms/configs/versions. However I'm not worried about that particular test file, that is for a very specific hardware/firmware test rig. We should however add some protocol/software level tests. On those I agree that we should maintain compatability.

PR #241, opened 2018-09-11
Fixed udev rules, udevadm trigger no longer required.

neomanic:
Because `/lib/udev/rules.d/50-udev-default.rules` lexically sorts after `50-odrive.rules`... our rule was getting overridden by the defaults. 🤦🏼‍♂️

Changed to `99-odrive.rules` to make sure it runs last.


kalvdans:
Looks ok to me!

madcowswe:
Great, this was also the solution I found [here](https://bbs.archlinux.org/viewtopic.php?pid=1520245#p1520245). Thanks for fixing it. I'll change it to 91 instead of 99 to give some extra space just in case.

PR #243, opened 2018-09-13
Analog input for position or velocity control

kalvdans:
This change makes it possible to control position or velocity with analog input signal. Fixes https://github.com/madcowswe/ODrive/issues/93 .

Example configuration:
```
odrv0.config.gpio3_analog_mapping.endpoint = odrv0.axis0.controller._remote_attributes['vel_setpoint']
odrv0.config.gpio3_analog_mapping.min = -400
odrv0.config.gpio3_analog_mapping.max = 400
```

### DIFF
--- a/Firmware/MotorControl/low_level.cpp
+++ b/Firmware/MotorControl/low_level.cpp
@@ -725,3 +725,32 @@ void pwm_in_cb(int channel, uint32_t timestamp) {
     last_pin_state[gpio_num - 1] = current_pin_state;
     last_sample_valid[gpio_num - 1] = true;
 }
+
+
+/* Analog speed control input */
+
+static void update_analog_endpoint(const struct PWMMapping_t *map, int gpio)
+{
+    float fraction = get_adc_voltage(get_gpio_port_by_pin(gpio), get_gpio_pin_by_pin(gpio)) / 3.3f;
+    float value = map->min + (fraction * (map->max - map->min));
+    get_endpoint(map->endpoint)->set_from_float(value);
+}
+
+static void analog_polling_thread(void *)
+{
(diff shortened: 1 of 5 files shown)


madcowswe:
Thanks! I want to consolidate our threads a bit and create a master "command handler" thread. Once we have that we should put this analog poll on that thread.
Do you mind leaving this PR hanging until we have that?
Thanks.

anilsawe:
Hi,
I made these changes to the firmware and was successful in driving the analog inputs to change the vel_setpoint. I am using two 9K ohm potentiometers to control the setpoints - connected to 3.3V (also tried Anlogg vcc).
However, with this firmware, my motors seemed to have stopped working. I am using Hoverboard motors and was successful in running them at various speeds for a long time - forward and reverse. Now the motors turn and stop. If I turn the potentiometer they will move forward and reverse but the motion stops after part of a revolution.

I went back to the original firmware and the problem persists. Is it possible that some settings have been corrupted? Hopefully, my ODrive board is not damaged. It's V3,4-24volts.

Any clues?

Anil Sawe
anilsawe@gmail.com

kalvdans:
@anilsawe I'm glad you tried the branch and got it working for a while. I suggest you dump all configuration variables and see if the operating mode has accidentally switched to sensorless.

@madcowswe Sorry for late reply, but I like the idea of having only one thread for the high-level control (the vesc firmware calls them apps).

madcowswe:
@kalvdans Yeah, it would be much easier to trace what could affect what is happening then currently. Nevertheless, I merged this now because it's still useful in it's current state, and we can refactor afterwards.
Thanks for your contribution!

PR #248, opened 2018-09-15
Addition of CMake Build System

SJ-Innovation:
CMake and CLion are excellent tools with debuggers that are very straightforward to use but extremely powerful. This PR adds the support of CMake and Clion to the build system portfolio, as well as beginning to split the system into a unified HAL instead of compiling only for F4.

This addition does not break the current VScode builder, VScode doesn't even know its there. The CMake builder has been designed so as to work around the VScode system, leaving no trace in any parts that VScode relies on. 

Configuration to the CMake builder is still done using the standard tup.config files used by VScode, basically if you can compile using VScode then you'll be able to compile with this too.  

Thanks for the great work on this project, and good things its bringing to the community! 

 - Sam :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ set(CMAKE_SYSTEM_NAME Generic)
 set(CMAKE_SYSTEM_VERSION 1)
 cmake_minimum_required(VERSION 3.7)
 
-# specify the cross compiler
+## Set for your ARM toolchain paths (or leave as is if you have them set in PATH variables)
 set(CMAKE_C_COMPILER_WORKS 1)
 set(CMAKE_C_COMPILER arm-none-eabi-gcc)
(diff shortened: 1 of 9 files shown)


SJ-Innovation:
Bumping for attention. Needy i know, but multiple redundant builders is important for distributed testing and debugging.

madcowswe:
Thanks for taking the time to write the CMake build script in a transparent way. However we are not interested in adding support for another build system at this time. Feel free to maintain a fork that carries this support.

If you can make a separate PR for just Abstracting the HAL config, I think we can accept that.

Thanks!

SJ-Innovation:
Won’t hide my disappointment ;) But no worries at all, though I’ll continue to hold my CMake fork open.

When I get a few spare minutes I’ll pull out the HAL abstractions and make a separate PR. 

I may also rework the CMake fork to directly use tup, and generate files from its outputs, as opposed to parsing the tupfiles directly.

PR #253, opened 2018-09-26
Property hooks

madcowswe:
Added post-write hooks to the protocol property implementation.
This allows hooking in updates for derived parameters. For example, when writing to the encoder bandwidth parameters, it can update the encoder gains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@ Please add a note of your changes below this heading if you make a Pull Request.
 
 # Unreleased
 
+### Added
+* Hook to execute protocol property written callback
+
+### Changed
+* Using new hooks to calculate:
+  * `motor.config.current_control_bandwidth`
+    * This depricates `motor.set_current_control_bandwidth()`
+  * `encoder.config.bandwidth`
+
 # Releases
 ## [0.4.4] - 2018-09-18
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Please add a note of your changes below this heading if you make a Pull Request.
   * Print function imported from future
 * Using new hooks to calculate:
   * `motor.config.current_control_bandwidth`
-    * This depricates `motor.set_current_control_bandwidth()`
+    * This deprecates `motor.set_current_control_bandwidth()`
   * `encoder.config.bandwidth`
 
 # Releases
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,15 @@ Please add a note of your changes below this heading if you make a Pull Request.
 
 # Unreleased
 
+### Added
+* Hook to execute protocol property written callback
+
+### Changed
+* Using new hooks to calculate:
+  * `motor.config.current_control_bandwidth`
+    * This depricates `motor.set_current_control_bandwidth()`
```
samuelsadok: depr**e**cates

Firmware/fibre/cpp/include/fibre/protocol.hpp
```diff
@@ -882,37 +883,48 @@ class ProtocolProperty : public Endpoint {
     }
     void handle(const uint8_t* input, size_t input_length, StreamSink* output) final {
         default_readwrite_endpoint_handler<TProperty>(property_, input, input_length, output);
+        if (written_hook_ != nullptr && input_length >= sizeof(TProperty)) {
```
samuelsadok: Although currently `input_length >= sizeof(TProperty)` is _functionally_ a necessary condition that the property was changed, _conceptually_ it is not. Meaning if you change the on-the-wire encoding this check might no longer be correct. It might be safer to have `default_readwrite_endpoint_handler` return a boolean and use that.

samuelsadok (approved):
Seems mostly good to me. Two general remarks:
 - This adds 8 bytes to every `ProtocolProperty` instance, and increases the stack size of where the whole thing is constructed. Be sure to check there is enough headroom (`system_stats`).
 - One potential danger is that if a developer updates the value from inside the firmware, he might expect that everything behaves the same and be surprised to find that the gains aren't updated. However this would be non-trivial to fix.

madcowswe:
Yep I checked the change in size of .bss section and checked the remaining stack size, both were acceptable IMO.
Thanks for reviewing this, I'll address the input_length based determination of written value.

Firmware/fibre/cpp/include/fibre/protocol.hpp
madcowswe: Yeah it felt dirty, but I didn't see another way to do it in this scope. The return value is a great idea, I'll do that instead.

PR #258, opened 2018-10-07
Implement basic CAN functionality

Wetmelon:


madcowswe:
some info here: https://discourse.odriverobotics.com/t/can-interface-available-for-testing/1448

Firmware/communication/interface_can.cpp
```diff
@@ -124,6 +124,7 @@ bool ODriveCAN::read(CAN_message_t &rxmsg) {
     rxmsg.isExt = header.IDE;
     rxmsg.id = rxmsg.isExt ? header.ExtId : header.StdId;  // If it's an extended message, pass the extended ID
     rxmsg.len = header.DLC;
+    rxmsg.rtr = (header.RTR == CAN_RTR_REMOTE) ? true : false;
```
kalvdans: The result of == is already a boolean.

Wetmelon: Lol I noticed that after I typed it but I was too lazy to change it :p

Firmware/communication/can_simple.cpp
```diff
@@ -0,0 +1,423 @@
...
+    axis->error_ |= Axis::ERROR_ESTOP_REQUESTED;
+}
+
+void CANSimple::get_motor_error_callback(Axis* axis, CAN_message_t& msg) {
+    if (msg.rtr) {
+        CAN_message_t txmsg;
+        txmsg.id = axis->config_.can_node_id << NUM_CMD_ID_BITS;
+        txmsg.id += MSG_GET_MOTOR_ERROR;  // heartbeat ID
+        txmsg.isExt = false;
+        txmsg.len = 8;
```
samuelsadok, line 141: This should probably be `txmsg.len = 4`.

Or maybe even `if (txmsg.len != 4) return;`. Based on [this explanation](http://www.esd-electronics-usa.com/CAN-Remote-Frames.html) collisions can occur if remote frame and data frame use different length fields so it may be good to discourage this by not responding, even if it sometimes works.

Wetmelon: I don't understand what you are referring to.

Wetmelon: We can use 4 bytes in this message if we want, but right now all messages are 8 bytes.  Common thing to do for whatever reason.

But why would I check for txmsg.len == 4 when I'm the one who is deciding how long the message is?

samuelsadok: Ah ok didn't know that was common.

The problem with remote frames is that if two nodes send the same remote frame but with a different length field, they can both win arbitration and then collide during the data field (I worded this a bit wrong before).

So it makes somewhat sense to encourage all client implementations to send the same data length field, by just ignoring any non-compliant remote frame. I mean once you received a non-compliant remote frame the collision was already evaded so you might as well just serve it, but in the long run the "ignore strategy" will kill off non-compliant implementations.

geg58:
Has anyone tested the communication rate achievable with CAN vs serial?

geg58:
Have you tested the ros package communication rate with the odrive while using CAN?

samuelsadok:
For the record: merged as part of #410

PR #260, opened 2018-10-10
Refactored HAL into more manageable and expandable format.

SJ-Innovation:
Hi!
I've refactored the HAL into a form that will be much more easily expanded in the future, i'm not currently finished with my work, as i still have yet to reorganise the USB/UART/GPIO settings to a relevant HAL module, but this should come fairly quickly.

Im also adding the ability to develop using other IDEs so we can hopefully increase our developer diversity. The builds are all done by exactly the same method (TUP) so there is no possibility of the results differing. My current plan for this is,

VSCode - Stable obviously (Original)
Eclipse - Usable (Original)
CLion - Dev, but stable so far through testing
Atom - Future... Perhaps? PlatformIO uses it with great success, may be worth looking into.

If you like what im doing feel free to make suggestions for improvements... If you hate what im doing.... Well what can I say?  xD

Kindest Regards,
Sam


Wetmelon:
This is kinda nice, much more general

SJ-Innovation:
Cheers! Im not finished with it yet, still wanting to pull most of the Hardware-Specific code out into a HAL, but once im done and satisfied ill send you a message.

metanoic:
> Atom - Future... Perhaps? PlatformIO uses it with great success, may be worth looking into.

I think until someone manages to get the board working with PlatformIO (would be very nice, IMO) that Atom isn't much of a concern.  I believe @Wetmelon or @Capo01 attempted and ran into some blockers with it.  I prefer Atom just 'cause I'm a webdev and am/was more familiar with it than VSCode.

With that said, also factor in the inevitable "I'm having problems with X and ya'll said it was possible" angle.

SJ-Innovation:
@metanoic Your are correct about there being a blocker, but it seems that the only thing that blocks is, is the lack of F405 support. I cant see that being too much of a challenge to add, considering we already have the linker and HAL files. PIO supports adding new boards, ill have a look into adding a specific ODrive board. Who knows... Might work xD

SJ-Innovation:
Okeidokie guys, Ive re-booted this little thread of mine with a few more changes. I'm currently at a state where I feel it is suitable enough to be merged in. If anyone has any suggestions or changes theyd like to see made before merging id be more tha  happy to discuss them and make edits if required. 

@Wetmelon @metanoic @madcowswe

Issue #103, opened 2018-01-20
Do control only every k PWM cycles


madcowswe:
We wish to boost the PWM frequency to get it out of the audible range. But the CPU can only run so fast.

madcowswe:
Note that we are unable to run low inductance motors until we complete this task, see [forum post](https://discourse.odriverobotics.com/t/error-brake-current-out-of-range-error-at-high-speeds/1182/12?u=madcowswe)

madcowswe:
Done!

PR #264, opened 2018-10-24
Increase PWM frequency

Wetmelon:
Run PWM at 24kHz and control loops at 8kHz.  Made default for all board variants.  Tested on a v3.3.  Odrive.ioc (CubeMX) file updated.

Addresses #103


madcowswe:
Thanks!

PR #271, opened 2018-11-14
Added RC PWM mixer for differential steering and PWM deadband.

jpwheless:
I've added a mixer that uses fibre endpoints for configuration.  Users can route the PWM signals to the mixer and then the mixer outputs to control parameters.

Here's the math inside the function:
endpoint_a = direction_a * (input_throttle + input_steering));
endpoint_b = direction_b * (input_throttle - input_steering));

I've tested it with a pistol-grip style transmitter and a hoverboard cart and I'm happy with the way it responds.

The PWM deadband feature adds a region from 1495μs to 1505μs where the value outputted is (min + max)/2.  Since this might not be useful for some applications, it has toggle to disable.

### DIFF
--- a/Firmware/MotorControl/low_level.cpp
+++ b/Firmware/MotorControl/low_level.cpp
@@ -679,30 +679,59 @@ void pwm_in_init() {
     }
 }
 
-#define TIM_2_5_CLOCK_HZ        TIM_APB1_CLOCK_HZ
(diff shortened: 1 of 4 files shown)


jpwheless:
Tested with RC PWM generator to confirm that min/max legal pulse time still works.  If the pulse time exceeds the min/max, the setpoint will stay at its last valid value, which was the behavior before this branch.

PR #274, opened 2018-11-28
Frequency domain anticogging map storage

TobinHall:
This implements storage of a finite number of frequency domain harmonics that describe the anticogging map.

The FFT / IFFT are perfomed using Arm CMSIS-DSP functions.

Currently only power series of two CPRs are supported.

This relies on knowing the absolute position of the motor.

### DIFF
--- a/Firmware/MotorControl/controller.cpp
+++ b/Firmware/MotorControl/controller.cpp
@@ -86,6 +86,7 @@ bool Controller::anticogging_calibration(float pos_estimate, float vel_estimate)
         } else {
             anticogging_.index = 0;
             set_pos_setpoint(0.0f, 0.0f, 0.0f);  // Send the motor home
+            extract_harmonics();
             anticogging_.use_anticogging = true;  // We're good to go, enable anti-cogging
             anticogging_.calib_anticogging = false;
             return true;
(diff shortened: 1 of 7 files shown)


Firmware/MotorControl/controller.cpp
```diff
@@ -85,13 +86,143 @@ bool Controller::anticogging_calibration(float pos_estimate, float vel_estimate)
...
+float Controller::write_harmonics(int32_t index, int32_t harmonic, int32_t imag, float value){
+    if (!std::isnan(value)){
+        config_.harmonics[index].index = harmonic;
+        imag ? config_.harmonics[index].imaginary = value : config_.harmonics[index].real = value;
+    }
+    return imag ? config_.harmonics[index].imaginary: config_.harmonics[index].real;
+}
+
+float Controller::write_anticogging_map(int32_t index, float value){
+    if (anticogging_.cogging_map != NULL && !std::isnan(value))
```
kalvdans, line 185: The NULL check here is useless since the array is read unconditionally below.

TobinHall: Good point. I have been a bit inconsistent with NULL checks of malloced arrays.

PAJohnson:
@TobinHall This PR is now incompatible with `devel`. Can you update or close it?

PR #275, opened 2018-11-28
enable bidirectional anticogging

Wetmelon:


Firmware/MotorControl/controller.cpp
```diff
@@ -75,19 +75,41 @@ void Controller::start_anticogging_calibration() {
...
+                }
+                break;
+
+            case ANTICOG_STATE_REV:
+                if (std::abs(pos_err) <= anticogging_.calib_pos_threshold &&
+                    std::abs(vel_estimate) < anticogging_.calib_vel_threshold) {
+                    anticogging_.cogging_map[anticogging_.index] += vel_integrator_current_;
+                    anticogging_.cogging_map[anticogging_.index--] /= 2.0f;
+                }
+                if (anticogging_.index > 0) {  // TODO: remove the dependency on encoder CPR
```
kalvdans: I don't know exactly how the anticogging mechanism works, but maybe this comparison should be `>= 0` so that all elements in cogging_map array are visited in the ANTICOG_STATE_REV state.

PR #276, opened 2018-11-28
Fix issues with spaces in install path on windows

ferret-guy:
This commit fixes issues on windows when you have spaces in the file path (i.e., when python3 gets installed to App Data under a user with a space in their name)


madcowswe:
Thanks!

PR #280, opened 2018-12-03
Input filter

madcowswe:
### Added
* Second order setpoint input filter.

### Changed
* Moved `traptraj.A_per_css` to `controller.inertia`
* Refactored velocity ramp mode into the new general input filtering structure

image


Wetmelon:
Given that we're doing everything through the input filter (passthrough, vel ramp, pos_filter, etc), I think we should make pos_setpoint and vel_setpoint read-only.  The commanded values are input_x and if you want one-to-one, use passthrough.  Might break users' scripts, but it feels odd to me that users can control the board so many ways (pos_setpoint, set_pos_setpoint(), input_pos).  Maybe we force them to use a function call instead?

EDIT:  Thinking about it, we should take this opportunity to pull in the TrajectoryPlanner into this same switch/case, and just use `move_to_pos()` or something along those lines for ALL moves.

madcowswe:
As noted there are too many ways to do the same thing:
```
<odrv>.<axis>.set_pos_setpoint()
<odrv>.<axis>.set_vel_setpoint()
<odrv>.<axis>.set_current_setpoint()
<odrv>.<axis>.move_to_pos()
<odrv>.<axis>.pos_setpoint = <value>
<odrv>.<axis>.vel_setpoint = <value>
<odrv>.<axis>.current_setpoint = <value>
<odrv>.<axis>.input_pos = <value>
<odrv>.<axis>.input_vel = <value>
<odrv>.<axis>.input_current = <value>
```

Agreed we can make the `xxx_setpoint` readonly, and kill the `set_xxx_setpoint()` functions: have the user only write to `input_xxx`. Agreed, we can merge the trajectory control mode into `INPUT_MODE_TRAP_TRAJ` and use a callback on writing to input_pos to call `move_to_pos`.

Wetmelon:
Updated as per the last note.

1. Made `xxx_setpoint` properties read-only
2. Removed `set_xxx_setpoint` and `move_to_pos` functions
3. Added `INPUT_MODE_TRAP_TRAJ`

Instead of a callback on input_pos, I added a simple check to see if the desired final position has changed.

Firmware/MotorControl/controller.cpp
```diff
@@ -55,7 +36,7 @@ void Controller::move_to_pos(float goal_point) {
                                  axis_->trap_.config_.accel_limit,
                                  axis_->trap_.config_.decel_limit);
     traj_start_loop_count_ = axis_->loop_counter_;
-    config_.control_mode = CTRL_MODE_TRAJECTORY_CONTROL;
+    trajectory_done_ = false;
```
Wetmelon, line 39: trajectory_done_ is never set true.  Is this intended?

PR #281, opened 2018-12-09
implement enable pin

madcowswe:
Instructions how to use [here](https://github.com/madcowswe/ODrive/blob/enable-pin/docs/interfaces.md#stepdirection)


Wetmelon:
This is now so far behind that I can't figure out how to merge it safely.  Please update.

emdash:
I want this functionality and would be interested in taking over this PR.

Firmware/Board/v3/Src/gpio.c
```diff
@@ -213,8 +213,10 @@ bool GPIO_subscribe(GPIO_TypeDef* GPIO_port, uint16_t GPIO_pin,
   struct subscription_t* subscription = NULL;
   for (size_t i = 0; i < n_subscriptions; ++i) {
     if (subscriptions[i].GPIO_port == GPIO_port &&
-        subscriptions[i].GPIO_pin == GPIO_pin)
+        subscriptions[i].GPIO_pin == GPIO_pin) {
       subscription = &subscriptions[i];
+      break;
```
samuelsadok, line 222: Not breaking on a match was intentional to aid in determinism

Firmware/MotorControl/axis.cpp
```diff
@@ -441,6 +477,7 @@ void Axis::run_state_machine_loop() {
                     task_chain_[pos++] = AXIS_STATE_CLOSED_LOOP_CONTROL;
                 else if (config_.startup_sensorless_control)
                     task_chain_[pos++] = AXIS_STATE_SENSORLESS_CONTROL;
+                task_chain_[pos++] = AXIS_STATE_STARTUP_SEQUENCE_DONE;
```
samuelsadok: This should be before closed loop control. It should also be added to `AXIS_STATE_FULL_CALIBRATION_SEQUENCE`.

Firmware/MotorControl/axis.cpp
```diff
@@ -521,6 +558,11 @@ void Axis::run_state_machine_loop() {
                 status = run_closed_loop_control_loop();
             } break;
 
+            case AXIS_STATE_STARTUP_SEQUENCE_DONE: {
+                startup_sequence_done_ = true;
+                status = true;
+            } break;
+
```
samuelsadok: In general the `AXIS_STATE_STARTUP_SEQUENCE_DONE` state doesn't feel quite right to me. It's not really a state. I suggest we set `startup_sequence_done_` the first time closed loop control is entered successfully.

Firmware/MotorControl/axis.hpp
```diff
@@ -59,6 +60,7 @@ class Axis {
     static LockinConfig_t default_lockin();
 
     struct Config_t {
+        bool startup_sequence_on_boot = false;   //<! run startup sequence on boot, run when asserting EN line otherwise
```
samuelsadok: There's a possible issue when this configuration option is `true`: If the enable pin is asserted and then deasserted during the startup sequence, the sequence will be aborted because idle state is requested on the deassert edge of EN. Next time EN is set, it attempts to go to closed loop control but will fail because things are not yet calibrated.

Is this an essential feature? Otherwise we can just leave it out and add it back carefully when we have a cleaner code base in general.

Firmware/MotorControl/axis.hpp
```diff
@@ -45,11 +47,14 @@ class Axis {
         bool startup_sensorless_control = false; //<! enable sensorless control after calibration/startup
         bool enable_step_dir = false; //<! enable step/dir input after calibration
                                     //   For M0 this has no effect if enable_uart is true
+        bool use_enable_pin = false;
+        bool enable_pin_active_low = true;
```
samuelsadok: Wouldn't a default of `false` be more expected by users?

madcowswe: Hm yes, lets change this.

Firmware/Board/v3/Src/gpio.c
madcowswe: Wouldn't that always give the _last_ match instead of first one? Is that good?
Also we don't really have very good timing determinism otherwise, and registering interrupts doesn't happen cyclically, so being fast is probably better.

samuelsadok: Fair point on the determinism, I left it like this now. The struct is zeroed out on unsubscribing so there should actually only ever be one matching item in the array.

annhan:
any update for this branch

PR #286, opened 2019-01-03
Update getting-started.md

mammat:
I was trying to work out how to  enable To enable Circular position control but could only find info on discord so added here


madcowswe:
Thanks!

Issue #230, opened 2018-09-01
0-100 PWM Input Mode


abickel:
Please create a PWM motor control function (0 to 100% duty cycle).
My specific application is spindle motor for a CNC Mill and the motion controller (Ethernet Smooth Stepper) outputs a 0~100% duty cycle at a user specified frequency.  In addition, the ODrive needs to accept a motor direction input (single pin).

Many thanks.

Wetmelon:
Per https://discourse.odriverobotics.com/t/velocity-control-using-pwm-input/1063/7

madcowswe:
Also known common mode:
* "0 to 100" -> "0 to max"
* "0 50 100" -> "-max 0 +max"

robsonswiss:
Just to clarify, "0 to 100" goes with a direction input as well.

Wetmelon:
Won't add, but we accept well-tested PRs.

Issue #262, opened 2018-10-21
Input processing module


madcowswe:
For direct input from a user, over say PWM, we may wish to modify the setpoints. For example:
* Stick mixing
* Deadband (to stop micro rolling)
* Filtering/interpolation (See also #110)

Wetmelon:
See also #280

xsherlockpl:
I'd love to see that happen to do really precise RC equipment control of the motors

Wetmelon:
We aren't adding any new large features to v3 firmware at this time, but we always accept a well-tested PR :)

PR #292, opened 2019-01-13
Duty-cycle PWM input, PWM deadband, differential steering mixer

jpwheless:
Adds duty-cycle PWM to the input capabilities of the ODrive. Used similarly to the existing RC PWM capabilities, this takes a square-wave signal with a duty cycle between 0% and 100% and a frequency between 25Hz and 500Hz. Direction pin capability was also added.

Adds a differential steering mixer that uses fibre endpoints for configuration. Users can route the PWM signals to the mixer and then the mixer outputs to control parameters.

Here's the math inside the function:
endpoint_a = direction_a * (input_throttle + input_steering));
endpoint_b = direction_b * (input_throttle - input_steering));

The PWM deadband feature adds a region from 1495μs to 1505μs (or 49.5% to 50.5%) where the value outputted is (min + max)/2. Since this might not be useful for some applications, it has toggle to disable.

New features tested on a hoverboard cart or with a PWM generator for functionality.

Fixes issue #230


jpwheless:
Also fixes some of issue #262

madcowswe:
Thank you! I will review this as soon as I am able to.

xsherlockpl:
Guys, can you make a deadband configurable in percent?. my RC car controller has like 2% return error  for throttle and I have a creep issues with my tracked GardenBot

emdash:
Is this abandoned?

emdash:
I would like to resurrect this feature if possible, but I see it's been in review for over a year. What do the ODrive devs want to see before this can be merged?

MatthewReed303:
When will this feature be added to devel or the master? What would be required in order to use 10khz PWM 0-100% rather than being limited to 500hz?

samuelsadok:
Sorry for letting this sit for so long. @madcowswe is quite busy with ODrive v4. I'll try to look at this in the next days. It won't make it into firmware v0.5.0 but hopefully v0.5.1.

@MatthewReed303 At a quick glance I don't see a hard limit at 500Hz but as you increase the frequency the PWM interrupt will occupy more and more CPU time, eventually causing the motor control timings to be missed (this would fail gracefully). So you can just try. Note that the motor control loop runs at 8kHz so you won't get faster response times than that.

MatthewReed303:
@samuelsadok  I'm using custom hardware I have made based on v3.5 but only using one axis. The main difference is I have 24v boost converter and jumpers for filters and jumpers for Encoder or Hall selection and dual-channel optocouplers with PNP transistors which can be set to inputs or outputs. I have just made it a little more compatible for interfacing to 24v systems/PLCs. Is there a way to fully disable the other axis so it's not using any resources at all? 

What will the difference be with v4?

samuelsadok:
Ah if it's single axis then there's a good chance it'll work with the CPU load.

M1 uses update interrupts of TIM8 so I believe you can disable that timer. It also uses and the ADC1 and ADC2 interrupts. However the ADCs are shared with M0. You'll want to disable the regular measurements and keep the injected measurements running. The relevant initialization code is in `low_level.cpp`, `tim.c` and `adc.c`.

ODrive v4 willl actually also be single axis and include I/O isolation and a bunch of other things.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+                pwm_dir_pin = true;
+            }
+        }
+        if (is_endpoint_ref_valid(board_config.pwm_mappings[gpio_num - 1].endpoint) || pwm_dir_pin) {
+            GPIO_InitStruct.Pin = get_gpio_pin_by_pin(gpio_num);
+            HAL_GPIO_DeInit(get_gpio_port_by_pin(gpio_num), get_gpio_pin_by_pin(gpio_num));
+            HAL_GPIO_Init(get_gpio_port_by_pin(gpio_num), &GPIO_InitStruct);
+            HAL_TIM_IC_ConfigChannel(&htim5, &sConfigIC, gpio_num_to_tim_2_5_channel(gpio_num));
+            HAL_TIM_IC_Start_IT(&htim5, gpio_num_to_tim_2_5_channel(gpio_num));
+        }
```
samuelsadok, line 139: You're initializing PWM direction pins as timer-capture pins instead of regular digital inputs. Using them as regular digital inputs would make more GPIOs eligible to be used as direction pins (all 8 GPIOs instead of only 4).

Is there a non-negligible chance that users want to
 - Use the same GPIO as PWM input _and_ direction input simultaneously
 - or have direction changes take effect faster than a PWM period?

MatthewReed303:
@samuelsadok I have been trying the devel merge into the Duty-cycle PWM you have done and can't get anything to work. Not even motor calibration and I get no error if I dump errors. I have tried debugging with vscode but don't know exactly where to start. If I go back to the devel branch and flash that everything works, so there is something wrong with the merge. Have you tried it yourself?

samuelsadok:
I didn't get to run it yet so it's quite possible that something went wrong with the merge. I'm gonna do a code review shortly and then also test it on hardware. If you happen to find the issue meanwhile let me know.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+            }
+        }
+    }
+
+    last_pin_state[gpio_index] = current_pin_state;
+
+    if (board_config.pwm_mappings[gpio_index].pwm_type == PWM_TYPE_DUTY_CYCLE) {
+        last_active_timestamp[gpio_index] = timestamp;
+        channel_active[gpio_index] = true;
+    }
```
samuelsadok, line 372: why is the active-detection for `PWM_TYPE_DUTY_CYCLE` and `PWM_TYPE_RC` in a different place?

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+            //   The interrupt function will set the channel active again if a valid pulse is received
+            // Duty-cycle PWM:
+            //   If the time since an edge exceeds PWM_INACTIVE_TIME, set channel inactive and set output to min or max depending on last_pin_state
+            //   Poll direction pin and set output correspondingly, because a change in that pin doesn't trigger an interrupt
+
+            switch(board_config.pwm_mappings[gpio_index].pwm_type) {
+                case PWM_TYPE_RC: {
+                    if ((htim5.Instance->CNT - last_active_timestamp[gpio_index]) > PWM_INACTIVE_TIME) {
+                        channel_active[gpio_index] = false;
+                        set_channel_output(endpoint, gpio_index, 0.5);
```
samuelsadok, line 401: I think we should error out if the PWM input is lost. Defaulting to 0.5 can be very dangerous. Imagine someone mapping RC input to velocity [0rpm, 1000rpm].

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+    
+#if HW_VERSION_MAJOR == 3 && HW_VERSION_MINOR >= 3
+    for (uint8_t gpio_index = 0; gpio_index < PWM_IN_COUNT; gpio_index++) {
+#else
+    // Only ch4 is available on v3.2
+    uint8_t gpio_index = 4; {
+#endif
+        // Only check channel if endpoint is assigned
+        Endpoint* endpoint = get_endpoint(board_config.pwm_mappings[gpio_index].endpoint);
+        if (endpoint && channel_active[gpio_index]) {
```
samuelsadok, line 386: ` && channel_active[gpio_index]` can be removed, sacrificing a handful of CPU cycles for less state.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+static uint32_t last_active_timestamp[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static uint32_t last_LtoH_timestamp[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static uint32_t last_HtoL_timestamp[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static uint32_t last_period_110[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static uint32_t last_period_090[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static uint32_t last_period[PWM_IN_COUNT] = { 0 }; // Timer ticks
+static bool last_pin_state[PWM_IN_COUNT] = { false };
+static bool channel_active[PWM_IN_COUNT] = { false };
+static float channel_value[PWM_IN_COUNT]; // Channel output value before modification by direction pin
```
samuelsadok, line 54: As far as I can see most of these state variables are redundant. Keeping just `last_LtoH_timestamp`, `last_HtoL_timestamp` and `last_pin_state` should be sufficient.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+// We could compare the period of each pulse to the previous one, and require one or more consecutive matches for the pulse to be considered valid.  This would
+// work for RC PWM too, and would replace the previous validation method.
+
+// dutyCycle = highTime / period
+
+// Calculating the period with either of these two formulas...
+// period = highTime + lowTime
+// period = HtoL_timestamp - last_HtoL_timestamp
+// ...depends on knowing the order of the high/low pulses in the cycle.  If we assume it's type two, it will be high then low.
+// As long as the duty cycle changes relatively slowly, if it's type 1 or 3 it shouldn't be too inaccurate.
```
samuelsadok, line 261: IMO hardcoding type 2 is fine. If another use case crops up we can change it later. In any case, we can assume that the PWM type is known.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+            HAL_TIM_IC_ConfigChannel(&htim5, &sConfigIC, gpio_num_to_tim_2_5_channel(gpio_num));
+            HAL_TIM_IC_Start_IT(&htim5, gpio_num_to_tim_2_5_channel(gpio_num));
+        }
+    }
+
+    // Create and start a periodic RTOS timer for pwm_no_pulse_check_cb()
+    osTimerDef_t pwm_check_callback_struct;
+    pwm_check_callback_struct.ptimer = &pwm_no_pulse_check_cb; // Callback function pointer
+    osTimerId pwm_check_timer = osTimerCreate(&pwm_check_callback_struct, osTimerPeriodic, NULL);
+    osTimerStart (pwm_check_timer, 5); // Period in milliseconds
```
samuelsadok, line 146: I think it would be better to set the timeout period to something like `max_pwm_period * 120%`. Otherwise there is a 5ms delay when jumping from >0% duty cycle to 0% duty cycle. I would be surprised at this as a user, more surprised than having to configure a nominal frequency.
We can for instance run this check on `tim_update_cb` (low_level.cpp) which runs at 16kHz (8kHz after the first `if`). That way we can keep separate periods for each PWM input and don't have to use FreeRTOS timers.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+// Calculating the period with either of these two formulas...
+// period = highTime + lowTime
+// period = HtoL_timestamp - last_HtoL_timestamp
+// ...depends on knowing the order of the high/low pulses in the cycle.  If we assume it's type two, it will be high then low.
+// As long as the duty cycle changes relatively slowly, if it's type 1 or 3 it shouldn't be too inaccurate.
+
+// We also need to handle the case where the PWM signal has no duty cycle, and is simply switching between max and min at random times.
+// We need to put a minimum period requirement on the signal, because there is no other way to differentiate between, say, switching from high to low
+// every 3 seconds, and a 50% PWM signal with a period of 6 seconds.
```
samuelsadok, line 265: We can ask the user to specify a nominal frequency or at least a maximum legal PWM period in order to properly detect 0% and 100% duty cycle.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+// Timer CCxOF bit can be checked to see if two captures occurred, the second overwriting the first
+// "In order to handle the overcapture, it is recommended to read the data before the
+//  overcapture flag. This is to avoid missing an overcapture which could happen after reading
+//  the flag and before reading the data."
+// Maybe we can use this to set the output values to min or max depending on current pin state
```
samuelsadok, line 271: The `CCxOF` bit would only help detecting a very short pulse, but not no pulse at all. Some form of timeout monitoring is required for that.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+// 3. The tail edge can be fixed and the lead edge modulated.
+
+// Is there a good way to check the period of all 3 types?  If not, type 2 should be the most common in this situation.
+// Type 2 is also the RC PWM signal type, so this implementation will work for both.
+
+// How can we verify that a pulse is a valid signal? RC PWM has this built-in, where a pulse needs to be within a certain length, and the period should be around
+// 20ms, but duty-cycle PWM has no such restrictions.  Right after boot, the old code waits for several HtoL transitions before it considers the input valid, but
+// shouldn't this be done every time a signal goes inactive?
+// We could compare the period of each pulse to the previous one, and require one or more consecutive matches for the pulse to be considered valid.  This would
+// work for RC PWM too, and would replace the previous validation method.
```
samuelsadok, line 253: If we ask the user to supply a nominal frequency, this becomes pretty straight forward: we can just check if the period is within a certain tolerance of the nominal period.
Regarding delay after start-up/bad signal, do we really need that? I don't remember/see such a thing in the old code.

Firmware/MotorControl/pwm_input.cpp
```diff
@@ -0,0 +1,423 @@
...
+    // Create and start a periodic RTOS timer for pwm_no_pulse_check_cb()
+    osTimerDef_t pwm_check_callback_struct;
+    pwm_check_callback_struct.ptimer = &pwm_no_pulse_check_cb; // Callback function pointer
+    osTimerId pwm_check_timer = osTimerCreate(&pwm_check_callback_struct, osTimerPeriodic, NULL);
+    osTimerStart (pwm_check_timer, 5); // Period in milliseconds
+}
+
+void diff_steering_mixer_update(int gpio_num) {
+    if (gpio_num != board_config.mixer_mapping.gpio_update_trigger)
+        return;
```
samuelsadok, line 151: For the sake of simplicity we should remove this check and update regardless.

samuelsadok (commented):
@jpwheless here are a couple of comments. I ran it against our RC PWM hardware tests and as far as I can tell that works fine as before. I didn't test the new duty cycle mode yet. Regarding the code I think the logic can be simplified quite a bit, see my inline comments. Please let us know if you'd like to address the comments or if you have lost interest and prefer someone else to take over.

Firmware/MotorControl/pwm_input.cpp
jpwheless: It looks like I did that to handle the case where duty-cycle PWM is at 100% or 0% and the direction pin changes.  Either it needed to be polled or a GPIO interrupt needed to be set up, because there would be no PWM edge to trigger a state check.  It wasn't to handle dual use as PWM and direction input.

jpwheless:
@samuelsadok That's some good feedback.  It would take me a while to find time and get a test environment set up again, so I'm okay with someone else taking over.  I'll keep looking at it, though.

Firmware/MotorControl/pwm_input.cpp
jpwheless: That's what I was using it for.  Timeout monitoring was being done in pwm_no_pulse_check_cb().

Firmware/MotorControl/pwm_input.cpp
jpwheless: You're right, having the user set a nominal PWM frequency instead of trying to detect it would get rid of a lot of this logic.

The timeout for jumping from >0% duty cycle to 0% duty cycle is actually PWM_INACTIVE_TIME, set to 45ms, so it's worse than you thought.  45ms works for RC PWM, but for duty-cycle PWM it should be set relative to the nominal frequency like you said.

Firmware/MotorControl/pwm_input.cpp
jpwheless: Agreed.

MatthewReed303:
@jpwheless @samuelsadok I can do some bench testing on the 0-100% PWM and report back with my findings. Can set up different frequency PWM.

 I think there should be an option to set what happens if PWM is lost. Either fault out and set the axis to idle or keep Odrive active and just set vel to 0 and hold pos in the current position. For example, Z-axis on a mill may be holding its own weight and if signal is lost for any reason we don't want to motor to just disable. Also same for a robot arm I'm planning to build for an injection molding machine where a braking system is difficult to implement. 

I think being able to use any pin for direction would be a better idea if possible. 

Another idea, Is velocity control via PWM without the closed-loop. For example, using a closed-loop CNC controller such as KFlop by DynoMotion. I would have thought 2x closed-loop systems fighting each other may cause issues. The Odrive will be trying to hold a velocity via PID and then the CNC controller is changing the velocity via PWM very quickly to maintain or drive to a position using its own PID loop. It may be fine, but I just don't know if the closed-loop velocity control can change at such rapid rates.

samuelsadok:
Yeah I suppose the most sensible lost-signal-action depends on the application and could be configurable to one of:

 - keep the last received value
 - fall back to a default value
 - error

Just to be clear though, this is only applicable to RC PWM. For the general duty-cycle-mode the ODrive can't really distinguish between a disconnected wire and 0%/100% duty cycle.

I believe you can't push to @jpwheless's branch so feel free to create a new PR if you make changes, then I'll close this one.

The onboard position controller is simply stacked onto the velocity controller. So by using a custom offboard position controller you'd get the same response times (plus the signal latency).

MatthewReed303:
I have done some testing and can't get the PWM to work at all. I have tried 500hz and 1khz verified by scope. No noise and a clean PWM signal. I will have an attempt at debugging. I noticed when putting it in closed loop it just stayed at 0 and the input_vel commands still worked even though I set PWM to be used, so it's almost like it did not even accept the PWM mode. Here is what I have done in order to try to get it to work. 

PWM 0-100% Control 
1: odrv0.config.gpio3_pwm_mapping.pwm_type = 1 // Set PWM to 0-100% | RC PWM Input = 0
2: odrv0.config.gpio3_pwm_mapping.min = 0  
3: odrv0.config.gpio3_pwm_mapping.max = 500000
4: odrv0.config.gpio3_pwm_mapping.endpoint = odrv0.axis0.controller._remote_attributes['input_vel']
5: odrv0.axis0.requested_state = AXIS_STATE_CLOSED_LOOP_CONTROL
6: odrv0.config.gpio3_pwm_mapping.endpoint = None // Disable PWM control

MatthewReed303:
@samuelsadok The debugging variables( valid_counts and all_counts ) just stay 0 so that indicates it's not even trying to count the PWM? 

I also have some other weird behavior going on. If I start Odrive it does its Index search fine. I check the status of the axis with odrv0.axis0 and the Input_vel and Input_pos is 0 including setpoint. When I set the controller into velocity mode and set to closed-loop control to motor just goes to full speed and I have no control. I go back and check the status of the axis again and some large random number has automatically be written to Input_pos and the setpoint. I have also disabled PWM with odrv0.config.gpio3_pwm_mapping.endpoint = None to make sure that's not causing any issues. Seems like the Odrive has a mind of its own and not responding to any input. Even if I set Input_pos or Input_vel to 0 while running nothing happens. I can set the axis back to Idle and set everything to 0 and put back into the closed-loop and the same thing happens but with a different random number written to Input_pos note Input_vel remains 0. What could cause this random behavior?

samuelsadok:
@MatthewReed303 that sounds odd.

Just so we're on the same page, can you try with this sequence:
```
odrv0.erase_configuration()
odrv0.axis0.requested_state = AXIS_STATE_MOTOR_CALIBRATION
[wait until end of beep]
odrv0.axis0.motor.config.pre_calibrated = True
odrv0.axis0.encoder.config.use_index = True
odrv0.axis0.requested_state = AXIS_STATE_ENCODER_INDEX_SEARCH
[wait until end of motion]
odrv0.axis0.requested_state = AXIS_STATE_ENCODER_OFFSET_CALIBRATION
[wait until end of motion]
odrv0.axis0.encoder.config.pre_calibrated = True
odrv0.axis0.controller.config.vel_limit_tolerance = 2 # for my setup the default is too low
odrv0.axis0.controller.config.control_mode = CONTROL_MODE_VELOCITY_CONTROL
odrv0.save_configuration()
odrv0.reboot()
odrv0.axis0.requested_state = AXIS_STATE_ENCODER_INDEX_SEARCH
[wait until end of motion]
odrv0.axis0.requested_state = AXIS_STATE_CLOSED_LOOP_CONTROL
odrv0.axis0.controller.input_vel = 20000
```

I just tried that with this branch (7cd48e5ab153b7a3e267778b3b24dffbf1072ab2) and it works fine on my setup.

`valid_counts` and `all_counts` are only incremented on actual pulses, not if duty cycle is 0% or 100%.

MatthewReed303:
@samuelsadok I have just tried your sequence and I'm getting an error invalid state when I put into closed loop mode

```
axis0
  axis: Error(s):
    ERROR_INVALID_STATE
  motor: no error
  encoder: no error
  controller: no error
```
Also, there is a typo for vel control mode that should be CTRL. I will investigate and see why I'm getting this error

I think it's because Encoder is not ready. I did note the encoder offset is only rotating one direction, I thought it goes both directions? 

If I go through my normal config procedure everything is fine.  
	MOTOR SETUP

	1: odrv0.axis0.encoder.config.cpr = 7200
	2: odrv0.config.brake_resistance = 1.5 
	3: odrv0.axis0.motor.config.pole_pairs = 4
	4: odrv0.axis0.motor.config.resistance_calib_max_voltage = 6
	5: odrv0.axis0.requested_state = AXIS_STATE_FULL_CALIBRATION_SEQUENCE
	6: odrv0.axis0.controller.config.vel_limit = 500000
	7: odrv0.axis0.motor.config.pre_calibrated = True
	8: odrv0.save_configuration()
	9: odrv0.reboot()

	ENCODER SETUP
	1: odrv0.axis0.encoder.config.use_index = True 
	2: odrv0.axis0.requested_state = AXIS_STATE_ENCODER_INDEX_SEARCH 
	3: odrv0.axis0.requested_state = AXIS_STATE_ENCODER_OFFSET_CALIBRATION 
	4: odrv0.axis0.encoder.config.offset  
	5: odrv0.axis0.encoder.config.pre_calibrated = True 
	6: odrv0.axis0.config.startup_encoder_index_search = True 
	7: odrv0.axis0.encoder.config.bandwidth = 2000 //change this if over 4000 cpr 
	8: odrv0.save_configuration()
	9: odrv0.reboot() 

	PID GAINS
	1: odrv0.axis0.controller.config.pos_gain = 10
	2: odrv0.axis0.controller.config.vel_gain = 0.0001 
	3: odrv0.axis0.controller.config.vel_integrator_gain = 0.015
	4: odrv0.save_configuration()
	5: odrv0.reboot()

	VELOCITY CONTROL
	1: odrv0.axis0.controller.config.control_mode = CTRL_MODE_VELOCITY_CONTROL 
	2: odrv0.axis0.requested_state = AXIS_STATE_CLOSED_LOOP_CONTROL
	3: odrv0.axis0.controller.input_vel = 10000
	4: odrv0.save_configuration()
	5: odrv0.reboot()

	I'm still getting the random position counts which seems odd. This config works just fine on the normal devel branch. I'm going to disconnect the PWM input and run through the procedure again. Here is the info of my axis0 if you pick up something odd. 

	error = 0x0000 (int)
	step_dir_active = False (bool)
	current_state = 1 (int)
	requested_state = 0 (int)
	loop_counter = 1129341 (int)
	lockin_state = 0 (int)
	is_homed = False (bool)
	config:
	  startup_motor_calibration = False (bool)
	  startup_encoder_index_search = True (bool)
	  startup_encoder_offset_calibration = False (bool)
	  startup_closed_loop_control = False (bool)
	  startup_sensorless_control = False (bool)
	  startup_homing = False (bool)
	  enable_step_dir = False (bool)
	  step_dir_always_on = False (bool)
	  counts_per_step = 2.0 (float)
	  watchdog_timeout = 0.0 (float)
	  enable_watchdog = False (bool)
	  step_gpio_pin = 1 (int)
	  dir_gpio_pin = 2 (int)
	  calibration_lockin: ...
	  sensorless_ramp: ...
	  general_lockin: ...
	  can_node_id = 0 (int)
	  can_node_id_extended = False (bool)
	  can_heartbeat_rate_ms = 100 (int)
	motor:
	  error = 0x0000 (int)
	  armed_state = 0 (int)
	  is_calibrated = True (bool)
	  current_meas_phB = 1.56499445438385 (float)
	  current_meas_phC = 1.295635461807251 (float)
	  DC_calib_phB = -2.7311549186706543 (float)
	  DC_calib_phC = -3.4324865341186523 (float)
	  phase_current_rev_gain = 0.02500000037252903 (float)
	  thermal_current_lim = 44.843650817871094 (float)
	  inverter_temp = 29.754039764404297 (float)
	  current_control: ...
	  gate_driver: ...
	  timing_log: ...
	  config: ...
	controller:
	  error = 0x0000 (int)
	  input_pos = 83978.4375 (float)
	  input_vel = 0.0 (float)
	  input_current = 0.0 (float)
	  pos_setpoint = 83978.4375 (float)
	  vel_setpoint = 0.0 (float)
	  current_setpoint = 0.0 (float)
	  trajectory_done = True (bool)
	  vel_integrator_current = 7.006492321624085e-44 (float)
	  anticogging_valid = False (bool)
	  gain_scheduling_width = 10.0 (float)
	  config: ...
	  move_incremental(displacement: float, from_goal_point: bool)
	  start_anticogging_calibration()
	encoder:
	  error = 0x0000 (int)
	  is_ready = True (bool)
	  index_found = True (bool)
	  shadow_count = -15793891 (int)
	  count_in_cpr = 2909 (int)
	  interpolation = 0.5 (float)
	  phase = -2.105794668197632 (float)
	  pos_estimate = -15793890.0 (float)
	  pos_cpr = 2909.43798828125 (float)
	  hall_state = 1 (int)
	  vel_estimate = 0.0 (float)
	  calib_scan_response = 0.0 (float)
	  pos_abs = 0 (int)
	  spi_error_rate = 0.0 (float)
	  config: ...
	  set_linear_count(count: int)
	sensorless_estimator:
	  error = 0x0000 (int)
	  phase = nan (float)
	  pll_pos = nan (float)
	  vel_estimate = nan (float)
	  config: ...
	trap_traj:
	  config: ...
	min_endstop:
	  endstop_state = False (bool)
	  config: ...
	max_endstop:
	  endstop_state = False (bool)
	  config: ...
	watchdog_feed()
	clear_errors()

	So with the PWM disconnected from gpio3 everything works as it should and how it was in the devel branch. So with PWM active on the Pin is causing something weird to happen. I there any way the PWM count could be writing to the input_pos? Here is the procedure I followed with the PWM connected after a full erase. 

	In [88]: odrv0.axis0.encoder.config.cpr = 7200                                                                   

	In [89]: odrv0.config.brake_resistance = 1.5                                                                     

	In [90]: odrv0.axis0.motor.config.pole_pairs = 4                                                                 

	In [91]: odrv0.axis0.motor.config.resistance_calib_max_voltage = 6                                               

	In [92]: odrv0.axis0.requested_state = AXIS_STATE_FULL_CALIBRATION_SEQUENCE                                      

	In [93]: odrv0.axis0.controller.config.vel_limit = 500000                                                        

	In [94]: odrv0.axis0.motor.config.pre_calibrated = True                                                          

	In [95]: odrv0.axis0.encoder.config.use_index = True                                                             

	In [96]: odrv0.axis0.requested_state = AXIS_STATE_ENCODER_INDEX_SEARCH                                           

	In [97]: odrv0.axis0.requested_state = AXIS_STATE_ENCODER_OFFSET_CALIBRATION                                     

	In [99]: odrv0.axis0.config.startup_encoder_index_search = True                                                  

	In [100]: odrv0.axis0.encoder.config.bandwidth = 2000                                                            

	In [101]: odrv0.axis0.controller.config.control_mode = CTRL_MODE_VELOCITY_CONTROL                                

	In [102]: odrv0.axis0.controller.config.pos_gain = 10                                                            

	In [103]: odrv0.axis0.controller.config.vel_gain = 0.0001                                                        

	In [104]: odrv0.axis0.controller.config.vel_integrator_gain = 0.015                                              

	In [105]: odrv0.save_configuration()                                                                             

	In [106]: odrv0.axis0.requested_state = AXIS_STATE_CLOSED_LOOP_CONTROL                                           

	In [107]: odrv0.axis0.controller.input_vel = 10000                                                               

	In [108]: odrv0.axis0.requested_state = AXIS_STATE_IDLE                                                          

	In [109]: odrv0.reboot()     

	After the reboot, everything is also fine, will do some more testing with PWM connected. So it seems I had something corrupt and embedded on the microcontroller... I did config with PWM connected and all the random issues, I ended up doing full chip erase via ST-Link and reflashed from clean and went through the following procedure again and now the PWM is working as it should. What could possibly get corrupt that a full chip erase is needed to get everything back in a working state? 

	When using the direction pin it works sometimes and other times I get an error...

	axis0
	  axis: Error(s):
		ERROR_MOTOR_DISARMED
	  motor: Error(s):
		ERROR_CONTROL_DEADLINE_MISSED
	  encoder: no error
	  controller: no error

Firmware/MotorControl/low_level.cpp
```diff
@@ -667,31 +667,59 @@ void pwm_in_init() {
...
+    if (channel != board_config.mixer_mapping.gpio_update_trigger)
+        return;
+
+    Endpoint* endpoint_a = get_endpoint(board_config.mixer_mapping.endpoint_output_a);
+    Endpoint* endpoint_b = get_endpoint(board_config.mixer_mapping.endpoint_output_b);
+    if (!endpoint_a || !endpoint_b)
+        return;
+    
+    endpoint_a->set_from_float(board_config.mixer_mapping.direction_a * (board_config.mixer_mapping.input_throttle + board_config.mixer_mapping.input_steering));
+    endpoint_b->set_from_float(board_config.mixer_mapping.direction_b * (board_config.mixer_mapping.input_throttle - board_config.mixer_mapping.input_steering));
```
Wetmelon: What happens when input_throttle and input_steering are both 100%?

PR #293, opened 2019-01-13
Update odrivetool.md, how to use dfu-util on macOS

jusic:
Add notes about using dfu-util on macOS. I successfully restored the firmware on macOS High Sierra to a ODrive 3.5 (48V) after a failed odrivetool dfu upload. The only caveat in these instructions is that I don't know any easy way to get objcopy on macOS, so I used a separate Ubuntu VM to convert the firmware .elf to a .bin.


madcowswe:
Thanks!

PR #300, opened 2019-01-16
Change getting-started.md with trajectory control

IamPhytan:
Trajectory Control commands weren't fully detailed.
Added command to set controller to `trajectory control`


madcowswe:
Thanks!

PR #301, opened 2019-01-18
Adding preliminary support for SPI absolute encoders.

TobinHall:
Adding preliminary support for non-blocking SPI absolute encoders. 
This code uses HAL_SPI_TransmitReceive_DMA() to start an SPI transaction. 
The transaction is initiated from the ADC callback and the SPI transaction complete interrupt calls the encoders callback.
The callback sets a flag which is checked in the encoder update function. This may not be necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,13 @@ tup.config
 /_site
 /.bundle
 
+
+ODrive\.config
+
+ODrive\.creator
+
+ODrive\.creator\.user
+
+ODrive\.files
+
+ODrive\.includes
--- a/Firmware/MotorControl/board_config_v3.h
+++ b/Firmware/MotorControl/board_config_v3.h
@@ -36,6 +36,7 @@ typedef struct {
     uint16_t hallB_pin;
     GPIO_TypeDef* hallC_port;
     uint16_t hallC_pin;
+    SPI_HandleTypeDef* spi;
 } EncoderHardwareConfig_t;
 typedef struct {
     TIM_HandleTypeDef* timer;
@@ -86,6 +87,7 @@ const BoardHardwareConfig_t hw_configs[2] = { {
         .hallB_pin = M0_ENC_B_Pin,
         .hallC_port = M0_ENC_Z_GPIO_Port,
         .hallC_pin = M0_ENC_Z_Pin,
+        .spi = &hspi3,
     },
     .motor_config = {
         .timer = &htim1,
@@ -125,6 +127,7 @@ const BoardHardwareConfig_t hw_configs[2] = { {
         .hallB_pin = M1_ENC_B_Pin,
         .hallC_port = M1_ENC_Z_GPIO_Port,
         .hallC_pin = M1_ENC_Z_Pin,
+        .spi = &hspi3,
     },
     .motor_config = {
         .timer = &htim8,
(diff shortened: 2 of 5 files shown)


Linjieqiang:
Can the TLE5012 encoder be supported?

TobinHall:
> Can the TLE5012 encoder be supported?

Possibly, I am not familiar with that encoder.

pluggee:
Hey @TobinHall, is there a list of encoders you're experimenting on?

TobinHall:
> Hey @TobinHall, is there a list of encoders you're experimenting on?

So far, I have only used the AS5047P but this should be common across most of the AS5xxx series I think. I will have an AMT223Bto test soon I believe.

adammck:
This works with the AS5048A, and rebases quite easily onto master. Nice work!

alireza0x0f:
can we use this branch to avoid startup encoder calibrations and read the current encoder position as startup position?

alireza0x0f:
@TobinHall thank you for your branch.
But i have a problem. I use a motor with 700 pole pairs and as5047p as encoder with 16384 cpr and when i start to calibrate encoder offset or encoder index search, motor and axis send error (ERROR_CONTROL_DEADLINE_MISSED and ERROR_MOTOR_DISARMED). If i set the pole pairs to 7 it is ok.
In addition i tested the 700 pole pairs motor with with incremental encoder (1000000 cpr) with master branch (fw 0.4.11) and it was ok.
it seems your branch is created from fw 0.4.7 and this problem was exist in that version.
Is there any way to use spi interface and as5047p encoder in new version of master branch?

TobinHall:
> @TobinHall thank you for your branch.
> But i have a problem. I use a motor with 700 pole pairs and as5047p as encoder with 16384 cpr and when i start to calibrate encoder offset or encoder index search, motor and axis send error (ERROR_CONTROL_DEADLINE_MISSED and ERROR_MOTOR_DISARMED). If i set the pole pairs to 7 it is ok.
> In addition i tested the 700 pole pairs motor with with incremental encoder (1000000 cpr) with master branch (fw 0.4.11) and it was ok.
> it seems your branch is created from fw 0.4.7 and this problem was exist in that version.
> Is there any way to use spi interface and as5047p encoder in new version of master branch?

@alireza0x0f I have not been able to spend any more time on supporting this feature. However, I believe @Wetmelon has this functionality added to his [RazorsEdge](https://github.com/Wetmelon/ODrive/tree/RazorsEdge) branch.

Nate711:
@TobinHall thanks for your work! Could you tell me what the startup and calibration procedure is like when using a spi encoder, as5047p for example? Thanks!

TobinHall:
> @TobinHall thanks for your work! Could you tell me what the startup and calibration procedure is like when using a spi encoder, as5047p for example? Thanks!

Hi Nate, 
It's been a while since I looked at this. Calibration should be the normal encoder offset calibration. This only needs to be performed once after which the encoder can be configured as precalibrated. 
In terms of startup, I believe I always manually moved into closed loop control. I never saw any issues on startup, but there could be a glitch as the encoder value is updated for the first time.

Hopefully that is of some help.
Tobin.

samuelsadok:
Thanks @TobinHall for this contribution!
This is now merged to `devel` as part of @Wetmelon's RazorsEdge branch (#410).

PR #311, opened 2019-02-07
PM Flux Linkage Measurement

adammunich:
Add flux linkage measurement to the motor calibration routine


Firmware/MotorControl/motor.hpp
```diff
@@ -50,6 +51,9 @@ class Motor {
         float overcurrent_trip_level; // [A]
     };
 
+    float cached_v_current_control_integral_d_;
+    float cached_v_current_control_integral_q_;
```
kalvdans: Why do these values need to be cached? It is not faster to access them here than under the `current_control_` struct.

adammunich: The values are best recorded at the very end of the steady-state part of the control loop, just before the motor goes into its freewheel state. If measured after the freewheel state then it is not guaranteed the measurements were not made under transient conditions.

kalvdans: I see, thanks for the explanation. I suggest removing "cached" from the name, since it is not possible to recalculate it on demand. It is more of a latched value then.

adammunich: I renamed it saved_steady_state_x which seemed more descriptive of its use.

Firmware/MotorControl/main.cpp
```diff
@@ -21,6 +21,11 @@ SystemStats_t system_stats_ = { 0 };
 
 Axis *axes[AXIS_COUNT];
 
+// Added for environments where new is not defined
```
SJ-Innovation, line 24: Bad idea on embedded. Why have you added this? I see nowhere that you're using it, nor should you use it given the extremely tight memory conditions of embedded systems.

adammunich: It would not compile for me without explicitly defining "new"

SJ-Innovation: You have issues with your builder/compiler then, what OS/tup version/arm gcc version are you using?

adammunich: Ubuntu 18, standard arm-none-eabi, using make, not tup.

Wetmelon: Yeah you have something wrong... malloc/new work fine.  What version of arm-none-eabi-gcc?  8.x is not supported.

adammunich: 7.3.0

adammunich:
Now compatible with lockin spin

PR #313, opened 2019-02-23
-

csann:
Revise Mac bash commands to install arm-gcc-bin library.


csann:
This edit fixes the Mac bash commands to install gcc-arm-bin. The commands in the original documentation were incorrect because the cask was removed from brew.

madcowswe:
Thanks!

PR #317, opened 2019-03-10
added control mode setters (current and velocity)

LuSeKa:
added functions in the Arduino library to set the ODrive control mode to either current or velocity

### DIFF
--- a/Arduino/ODriveArduino/ODriveArduino.cpp
+++ b/Arduino/ODriveArduino/ODriveArduino.cpp
@@ -46,6 +46,14 @@ void ODriveArduino::TrapezoidalMove(int motor_number, float position){
     serial_ << "t " << motor_number << " " << position << "\n";
 }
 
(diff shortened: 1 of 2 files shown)


madcowswe:
Thanks for making this.
What is the situation when you need to change the control mode, but you don't want to call SetVelocity or SetPosition to change into vel or pos mode?

LuSeKa:
In my application (mobile robot control) I need the Arduino to change the ODrive's control mode between current and velocity. 
Does issuing SetVelocity() and SetCurrent() automatically change between these control modes? I might have overlooked this. 
If so, then separate setters for the control modes are probably unneccesary.

Wetmelon:
> Does issuing SetVelocity() and SetCurrent() automatically change between these control modes? I might have overlooked this.

Yes, internal to the ODrive

LuSeKa:
OK cool, I'll test

PR #322, opened 2019-03-31
Added more information on analog inputs

Capo01:
I have just been reading over the docs and I have noticed that there is not much on the topic of using the GPIO as analog inputs beyond stating what pins are available.


madcowswe:
Thanks!

PR #324, opened 2019-04-02
Adam made these changes a while ago and this should merge with your master

jerkey:
NOTE:  adam's work was based on branch "devel" so this PR may be inappropriate.  BRB, making another

these changes let you measure flux linkage!  very impressive.

automatically merged with madcowswe/ODrive:master today


madcowswe:
Duplicate of #323

PR #325, opened 2019-04-02
Adam added flux linkage measurement

jerkey:
these changes let you measure flux linkage!  very impressive.

automatically merged with madcowswe/ODrive:devel today


madcowswe:
Duplicate of #323

PR #327, opened 2019-04-02
Fiber JSON cache

adammunich:
Caching JSON for fibre, compare crc to invalidate. Makes reconnecting faster.

### DIFF
--- a/Firmware/fibre/python/fibre/discovery.py
+++ b/Firmware/fibre/python/fibre/discovery.py
@@ -12,6 +12,8 @@ import fibre.utils
 import fibre.remote_object
 from fibre.utils import Event, Logger
 from fibre.protocol import ChannelBrokenException, TimeoutError
+import tempfile
+import os
 
 # Load all installed transport layers
 
@@ -41,6 +43,7 @@ try:
 except ImportError:
     pass
 
+
 def noprint(text):
     pass
 
@@ -54,7 +57,6 @@ def find_all(path, serial_number,
     the callback for each Fibre node that is found.
     This function is non-blocking.
     """
-
     def did_discover_channel(channel):
         """
         Inits an object from a given channel and then calls did_discover_object_callback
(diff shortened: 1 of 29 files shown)


Firmware/fibre/python/fibre/discovery.py
```diff
@@ -64,25 +67,71 @@ def did_discover_channel(channel):
...
+
+            cache_path =  temp_dir + '/fibre_schema_cache_' + str(json_crc16)
+
+            try:
+                json_cache = open(cache_path, 'r+')
+                logger.debug("Found cache file with crc")
+                cache_miss = False
+            except:
+                logger.debug("Cache miss, there is no cache file")
+                cache_miss = True
```
madcowswe: Can you merge this try block with the one above, and hence get rid of json_crc16 = 0 ?
If you want to have a default for old firmware, you need to put it in the except block, or it may be ill-formated.

Firmware/fibre/python/fibre/discovery.py
```diff
@@ -54,7 +58,6 @@ def find_all(path, serial_number,
     the callback for each Fibre node that is found.
     This function is non-blocking.
     """
-
```
madcowswe, line 57: These random line changes, do they have a reason?

Firmware/fibre/python/fibre/discovery.py
```diff
@@ -64,25 +67,71 @@ def did_discover_channel(channel):
...
+                    json_cache = open(cache_path, 'w+')
+                    json_cache.write(json_bytes.decode("ascii"))
+                    logger.debug("saved json_bytes to file")
+
+                except (TimeoutError, ChannelBrokenException):
+                    logger.debug("no response - probably incompatible")
+                    return
+            else:
+                try:
+                    json_bytes = json_cache.read().encode("ascii")
```
madcowswe: This seems redundant? Either store the ascii bytes, or store the decoded string. No point in ever going string -> bytes.

adammunich: I got an error when trying to use the file functions with bytes, it expects a string with encoding

madcowswe: okay, just store the string, and arrive at json_string directly (no need to reconstruct some json_bytes)

samuelsadok:
The more up-to-date version of this (#353) has been merged to `RazorsFrozenTesting` which in turn will be merged into devel soon, so I'm gonna go ahead and close this one.

PR #329, opened 2019-04-11
Corrected the native protocol stream format docs

DomenicP:
The packet length field is currently only one byte, not two.


madcowswe:
Thanks!

PR #335, opened 2019-05-04
removed f from gain odrive commands

Capo01:
When run in odrivetool the current commands will give a syntax error.


madcowswe:
Thanks!

PR #336, opened 2019-05-05
Included description of * `ERROR_OVERSPEED = 0x01`

Capo01:
I have been encountering this error when running two motors back to back and this is my current understanding of why it is caused.

Just a side note: Perhaps it may be useful for this error to only trigger if `vel_limit` is exceeded for some meaningful length of time. e.g. > 250 ms.


madcowswe:
Thanks!

PR #337, opened 2019-05-05
Mention low RPM, high braking torque ERROR_DC_BUS_OVER_VOLTAGE edge case

Capo01:
While testing a motor dyno I encountered an issue where energy could not be absorbed by the brake resistor at low RPM due to the small voltage generated by the motor. This took me a little wire to figure out and so I have included mentioned it here.

I have not made any suggestions for how to resolve the issue as this will depend on the circumstances of each user. Perhaps if there is a blog post on the forum describing a similar issue that could be linked here.


madcowswe:
I don't think this is how the brake resistor circuit works. We can discuss more on discord.

Issue #207, opened 2018-08-04
Reliability issues on ASCII protocol on USB


madcowswe:
Sending position commands at a high rate stalls after some time. See [original thread](https://discourse.odriverobotics.com/t/ascii-commands-makes-odrive-unreachable/849), which includes a test program for windows.

madcowswe:
Problems also reported in [this thread](https://discourse.odriverobotics.com/t/serial-timeout-issue/1039).

rklosinski:
I have the same problem. Hangs with multiple serial commands. How close is this to being corrected?

27183:
We are running the binary protocol and having some commands contain embedded bytes that can be interpreted as valid ASCII commands; which are not intended and do damage.

Wetmelon:
Closing until we hear anything new from 0.5.3.

PR #340, opened 2019-05-11
Fix buffer overflow issue in 'w' and 'r' commands

Wetmelon:
We can only store a string that is 1 character smaller than the destination string due to the null-terminating character '\0'

Found with CppCheck static analysis.

Potential fix for #207


kalvdans (commented):
I think it is better to change the allocation to `char name[MAX_LINE_LENGTH + 1];` so we don't lose the beauty of having the length definition in one place.

Wetmelon:
Agreed, as long as there's no issue with the string being > 256 bytes.

PR #351, opened 2019-05-25
added index_search and noise reduction information

owhite:
Added some text on a method to test if index_search is working, and some mitigating tactics to remove noise on the Z-input.


madcowswe:
Thanks!

PR #353, opened 2019-05-28
Fiber cache

kingoflolz:
caches fiber json by crc, also persists across reboots now

### DIFF
--- a/Firmware/fibre/python/fibre/discovery.py
+++ b/Firmware/fibre/python/fibre/discovery.py
@@ -13,7 +13,7 @@ import fibre.utils
 import fibre.remote_object
 from fibre.utils import Event, Logger
 from fibre.protocol import ChannelBrokenException, TimeoutError
-import tempfile
+import appdirs
 import os
 
 # Load all installed transport layers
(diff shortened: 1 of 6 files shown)


Firmware/fibre/python/fibre/discovery.py
```diff
@@ -68,75 +94,52 @@ def did_discover_channel(channel):
...
-                    return  
+                    raise UnicodeDecodeError
+
                 json_crc16 = fibre.protocol.calc_crc16(fibre.protocol.PROTOCOL_VERSION, json_bytes)
 
+                cache_path = temp_dir + '/fibre_schema_cache_' + str(json_crc16)
+
+                logger.debug(f"Opening json cache file {cache_path}")
+                json_cache = open(cache_path, 'w+')
+                json_cache.write(json_string)
```
madcowswe: This file is never closed. Perhaps use "with open" syntax?

madcowswe (commented):
Can you also check if:
1. old odrivetool still works with new firmware
2. new odrivetool still works with old firmware

adammunich:
see https://github.com/madcowswe/ODrive/pull/358

PR #354, opened 2019-05-29
Remove stack pointer from clobber list

samuelsadok:
This fixes an issue when compiling with GCC 9.1:
"In function 'void early_start_checks()': error: listing the stack pointer register 'sp' in a clobber list is deprecated [-Werror=deprecated]"

Corresponds fix in upstream ARM CMSIS:
https://github.com/ARM-software/CMSIS_5/commit/a07b56505471fd789111bd7dba726461449cc261

Upgrading the whole CMSIS would be way more involved so I just PR this quick fix here.


madcowswe:
Thanks!

PR #355, opened 2019-05-29
refactor lockin mode

kingoflolz:
refactor lockin mode into separate variables for sensorless spinup, encoder calibration and general (user initiated) mode

### DIFF
--- a/Firmware/MotorControl/axis.cpp
+++ b/Firmware/MotorControl/axis.cpp
@@ -33,6 +33,48 @@ Axis::Axis(int axis_num,
     update_watchdog_settings();
 }
 
+Axis::LockinConfig_t Axis::default_calibration() {
+    Axis::LockinConfig_t config;
+    config.current = 10.0f;           // [A]
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 20.0f;     // [rad/s^2]
+    config.vel = 40.0f; // [rad/s]
+    config.finish_distance = 100.0f * 2.0f * M_PI;  // [rad]
+    config.finish_on_vel = false;
+    config.finish_on_distance = true;
+    config.finish_on_enc_idx = true;
+    return config;
+}
+
+Axis::LockinConfig_t Axis::default_sensorless() {
+    Axis::LockinConfig_t config;
+    config.current = 10.0f;           // [A]
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 200.0f;     // [rad/s^2]
+    config.vel = 400.0f; // [rad/s]
+    config.finish_distance = 100.0f;  // [rad]
+    config.finish_on_vel = true;
+    config.finish_on_distance = false;
+    config.finish_on_enc_idx = false;
+    return config;
+}
+
+Axis::LockinConfig_t Axis::default_lockin() {
+    Axis::LockinConfig_t config;
+    config.current = 10.0f;           // [A]
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 20.0f;     // [rad/s^2]
+    config.vel = 400.0f; // [rad/s]
+    config.finish_distance = 100.0f;  // [rad]
+    config.finish_on_vel = false;
+    config.finish_on_distance = false;
+    config.finish_on_enc_idx = false;
+    return config;
+}
+
 static void step_cb_wrapper(void* ctx) {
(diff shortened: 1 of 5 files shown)


Firmware/MotorControl/axis.hpp
```diff
@@ -49,6 +49,10 @@ class Axis {
         bool finish_on_enc_idx = false;
     };
 
+    static LockinConfig_t default_calibration();
+    static LockinConfig_t default_sensorless();
+    static LockinConfig_t default_lockin();
```
madcowswe, line 54: What is the purpose of these statics?

Firmware/MotorControl/axis.cpp
```diff
@@ -179,44 +221,44 @@ bool Axis::watchdog_check() {
     }
 }
 
-bool Axis::run_lockin_spin() {
+bool Axis::run_lockin_spin(LockinConfig_t &lockin_config) {
```
madcowswe: make reference "const" if you don't intend to change the value

Firmware/MotorControl/axis.hpp
```diff
@@ -240,7 +246,25 @@ class Axis {
                     [](void* ctx) { static_cast<Axis*>(ctx)->decode_step_dir_pins(); }, this),
                 make_protocol_property("dir_gpio_pin", &config_.dir_gpio_pin,
                     [](void* ctx) { static_cast<Axis*>(ctx)->decode_step_dir_pins(); }, this),
-                make_protocol_object("lockin",
+                make_protocol_object("calibration",
```
madcowswe: The word "lockin" should still exist here, otherwise it's not clear to the user what kind of config this is.

madcowswe: Maybe this should be true of the struct name also, `config_.calibration.current` doesn't make it clear its' about lockin settings.

Firmware/MotorControl/axis.cpp
```diff
@@ -33,6 +33,48 @@ Axis::Axis(int axis_num,
...
+    config.finish_on_enc_idx = true;
+    return config;
+}
+
+Axis::LockinConfig_t Axis::default_sensorless() {
+    Axis::LockinConfig_t config;
+    config.current = 10.0f;           // [A]
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 20.0f;     // [rad/s^2]
```
madcowswe: This accel is very slow for sensorless spinup, it should probably get to full speed in 0.5-2s. Did you try sensorless yet?

Firmware/MotorControl/axis.cpp
```diff
@@ -33,6 +33,48 @@ Axis::Axis(int axis_num,
...
+Axis::LockinConfig_t Axis::default_calibration() {
+    Axis::LockinConfig_t config;
+    config.current = 10.0f;           // [A]
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 20.0f;     // [rad/s^2]
+    config.vel = 40.0f; // [rad/s]
+    config.finish_distance = 100.0f;  // [rad]
+    config.finish_on_vel = false;
+    config.finish_on_distance = true;
```
madcowswe, line 45: If you are going to have finish on distance by default, you should increase the distance to be more than the max likely pole pairs we will ever see. Like 50pp? so like finish_distance = 2pi*50, something like that?

Firmware/MotorControl/axis.hpp
kingoflolz: cpp doesn't let you use them in default constructors if its not static

Firmware/MotorControl/axis.hpp
kingoflolz: changed to calibration_lockin and sensorless_ramp

Firmware/MotorControl/axis.cpp
```diff
@@ -33,6 +33,48 @@ Axis::Axis(int axis_num,
...
+    config.ramp_time = 0.4f;          // [s]
+    config.ramp_distance = 1 * M_PI;  // [rad]
+    config.accel = 20.0f;     // [rad/s^2]
+    config.vel = 400.0f; // [rad/s]
+    config.finish_distance = 100.0f;  // [rad]
+    config.finish_on_vel = false;
+    config.finish_on_distance = false;
+    config.finish_on_enc_idx = false;
+    return config;
+}
```
madcowswe: To avoid duplication, maybe omit the default function, and instead just use the values set in the definition of LockinConfig_t  in axis.hpp. To restore those defaults, we would just use the default constructor.

kingoflolz: think I did what you want? not 100% sure if I understood what you meant

PR #357, opened 2019-06-13
Velocity Limiting in CURRENT_CONTROL_MODE

yconst:
Adds velocity limiting in current control mode, similar to what is described [here](https://discourse.odriverobotics.com/t/limiting-velocity-in-current-control-mode/485). ~~This is draft as it hasn't been tested yet. Tests pending.~~


yconst:
This has been tested on M0 of an ODrive 3.5 with 0.4.10 and performs as expected.

In current control mode and while `vel_limit > 0` and `vel_gain > 0`, current is adjusted so as to avoid exceeding the `vel_limit`. Marking this as ready for review.

madcowswe:
Thanks!

Wetmelon:
Sorry, this doesn't pass my unit testing.  Please describe the expected functionality fully.

Consider the case when:

vel_limit = 1000
vel_estimate = 999
vel_gain = 1
Iq = -5
Imax is (1000 - 999) * 1 = 1

The limited Iq = -1, which doesn't make any sense.  The motor is trying to decelerate and we're limiting  that behaviour.  In fact, if vel_estimate == vel_limit then the current controller does nothing.  It will be incapable of slowing the motor down until friction naturally slows it down and increases the velocity error.


image


My expectation is that the current will be clamped between [-current_lim, Imax] when vel_estimate > 0 and [Imax, current_lim] when vel_estimate < 0

yconst:
You're right about opposing `Iq` being erroneously limited as well. I think a check for `vel_estimate > 0.` in addition to the other checks should fix that. I'll write it up and update the PR.

About the clamping range, I'm wondering if it's desired behavior to actively limiting overspeed, in other words, if `vel_estimate > vel_limit`, should it be that `Iq < 0` ? or should it be` Iq == 0` ? (similarly for negative current and vel) My opinion would be towards the latter, as otherwise `vel_limit` functions as a kind of setpoint. But it's a design choice I think

yconst:
btw, is the test suite that you use available?

Wetmelon:
Yeah, I use doctest.

turns out that vel_estimate checking is a crappy way to do this.  Let me futz around with it and see what that parallelogram shape ends up being mathematically

New algorithm does not do any checking for vel_estimate.  It is just calculating the Y point of two lines, where `Imax = vel_gain * x + vel_limit` and `Imin = vel_gain * x - vel_limit` and then clamping within the current_lim and the lines.

```C++
float limitVel(const float vel_limit, const float vel_estimate, const float vel_gain, const float Iq, const float current_lim) {
    float Imax = std::min((vel_limit - vel_estimate) * vel_gain, current_lim);
    float Imin = std::max((-vel_limit - vel_estimate) * vel_gain, -current_lim);
    return std::clamp(Iq, Imin, Imax);
}
```

yconst:
Current limiting is already done [further on in the code](https://github.com/madcowswe/ODrive/blob/7e0977a73aba6e70f53a0e9d9db471f3265265d7/Firmware/MotorControl/controller.cpp#L195).. so what's the use of it here?

Also this does not set the `limited` flag, if that's important..

There's also a chance that `Imax < Imin` (when `vel_estimate > vel_limit + current_lim/vel_gain` ), [which leads to undefined behavior](https://en.cppreference.com/w/cpp/algorithm/clamp)

Wetmelon:
Quite right, sir.  I think this will always be right (assuming vel_limit and vel_gain are both positive).  `limited` flag is only for the vel_integrator_gain which isn't used in current mode.

```C++
float limitVel(const float vel_limit, const float vel_estimate, const float vel_gain, const float Iq, const float current_lim) {
    float Imax = (vel_limit - vel_estimate) * vel_gain;
    float Imin = (-vel_limit - vel_estimate) * vel_gain;
    return std::clamp(Iq, Imin, Imax);
}
```

yconst:
Yes, this seems to do it.

Also, It may be the case that `std::clamp(Iq, Imin, Imax) == Iq` so in this case current is not limited, but since the `limited` flag is not relevant it's not important.

yconst:
A small suggestion having just seen your latest implementation: In [L255](https://github.com/Wetmelon/ODrive/blob/1494cb333fa4e087057f1b2cb9320c28062bd51d/Firmware/MotorControl/controller.cpp#L255), instead of checking for `config_.vel_limit > 0.0f`, I think it would be better to have a separate flag that enables/disables velocity limiter, as users may wish to have `vel_limit == 0` and have the ODrive braking. This was suggested by @madcowswe in the forum.

PR #358, opened 2019-06-20
Fiber cache with more entropy

adammunich:
This bumps the 1/65,535 chance of having a cache collision to 2^32 or thereabouts, (disregarding the non-uniformity distribution of CRC... so it's probably actually 2^31)

Also fixed broken exceptionhandler in shell.py

### DIFF
--- a/Firmware/fibre/python/fibre/discovery.py
+++ b/Firmware/fibre/python/fibre/discovery.py
@@ -13,7 +13,7 @@ import fibre.utils
 import fibre.remote_object
 from fibre.utils import Event, Logger
 from fibre.protocol import ChannelBrokenException, TimeoutError
-import tempfile
+import appdirs
 import os
 
 # Load all installed transport layers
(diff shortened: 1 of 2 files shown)


Firmware/fibre/cpp/include/fibre/protocol.hpp
```diff
@@ -1,3 +1,5 @@
+#pragma GCC optimize ("O3")
```
samuelsadok: I think placing this here would apply this setting to most of the firmware, so we should probably discuss it separately from this PR

Firmware/fibre/cpp/protocol.cpp
```diff
@@ -187,8 +188,8 @@ int BidirectionalPacketBasedChannel::process_packet(const uint8_t* buffer, size_
         // Verify packet trailer. The expected trailer value depends on the selected endpoint.
         // For endpoint 0 this is just the protocol version, for all other endpoints it's a
         // CRC over the entire JSON descriptor tree (this may change in future versions).
-        uint16_t expected_trailer = endpoint_id ? json_crc_ : PROTOCOL_VERSION;
-        uint16_t actual_trailer = buffer[length - 2] | (buffer[length - 1] << 8);
+        uint32_t expected_trailer = endpoint_id ? json_crc_ : PROTOCOL_VERSION;
+        uint32_t actual_trailer = buffer[length - 2] | (buffer[length - 1] << 8);
```
samuelsadok: this looks like a spurious change

Firmware/fibre/python/fibre/discovery.py
```diff
@@ -122,23 +124,25 @@ def did_discover_channel(channel):
                     raise UnicodeDecodeError
 
                 json_crc16 = fibre.protocol.calc_crc16(fibre.protocol.PROTOCOL_VERSION, json_bytes)
-
-                cache_path = temp_dir + '/fibre_schema_cache_' + str(json_crc16)
-
-                logger.debug(f"Opening json cache file {cache_path}")
+                json_fibre_cache_entropy = fibre.protocol.calc_crc16(json_crc16, json_bytes) | (json_crc16 << 16)
```
samuelsadok: IMO we should treat `json_fibre_cache_entropy` as an opaque cache file ID, i.e. fetch it from the device here and not derive it by calculation. Doing so leaves us free to change how it's calculated at a later point (e.g. a CRC32).

PR #362, opened 2019-07-05
Intial brushed motor control working.

sequoia-hope:
This is a clean commit of the minimum changes I made to get brushed motor control working. This is voltage control only, using the "current" command value as Volts.

Here's the Python necessary to test this:

odrv0.axis0.motor.config.direction = 1
odrv0.axis0.controller.config.control_mode = CTRL_MODE_CURRENT_CONTROL
odrv0.axis0.motor.config.motor_type = MOTOR_TYPE_BRUSHED_VOLTAGE
odrv0.axis0.requested_state = AXIS_STATE_BRUSHED_VOLTAGE_CONTROL
odrv0.axis0.controller.current_setpoint = 10  #Will be 10 Volts


Wetmelon:
Awesome, thank you!  Reviews / PR merges are in limbo for now but we'll let you know as soon as there's any movement.

rowanG077:
Will this be picked up soon?

samuelsadok:
Now that #472 is merged I think the code base is in a decent position to accommodate something like this. But currently we're quite busy with higher priority tasks so someone else would have to update it (@tlalexander?).

The brushed motor current/voltage controller would need to be refactored into a class that inherits from [PhaseControlLaw](https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/phase_control_law.hpp). It would then serve as a alternative to [FieldOrientedController](https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/foc.cpp). It would take a current and/or voltage setpoint as an input and control two of the three phases accordingly. I guess you could call it DcController or so. The [Motor::update() function](https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/motor.cpp#L496) would need to be modified to convert a torque input to a current or voltage setpoint for the new DC control law.

sequoia-hope:
Thanks @samuelsadok for the highlights on the necessary changes. Yes I've noticed the big changes and that my branch has fallen behind. I am happy to take on the updates when time permits, but I am not sure when that will be at the moment. I will keep y'all posted if I find the time.

Wetmelon:
Hi @tlalexander I think at this point we're going to call this closed until you update the PR.  We don't expect to change anything else major after 0.5.3, so it's a good time to update the PR.  In the meantime, if anyone asks, we will point them to your branch / fork.

PR #365, opened 2019-07-20
Update commands.md. set motor.config.direction

Launchpad-McQ:
need to set motor.config.direction to something other than 0 (0 is the default, maybe change?) or state wont change to AXIS_STATE_SENSORLESS_CONTROL.


madcowswe:
Thanks!

PR #375, opened 2019-09-19
fix bad use of memcpy

samuelsadok:
memcpy is not guaranteed to work with overlapped memory areas, however memmove is.
Apart from the semantic difference this fixes a compiler warning and possibly previously unexplained occurrences of `ERROR_INVALID_STATE`.


kalvdans:
Looks good to me!

PR #378, opened 2019-09-27
support "unsigned int" type in native protocol (aka Fibre)

samuelsadok:
This fixes a compile error when using enums with large numerical values. The int type of such enums is unsigned int, which was previously not supported.


madcowswe:
Thanks

PR #383, opened 2019-10-19
Dynamically reconfigure UART to GPIO 3 and 4

tobbelobb:
### WIP: Only tested to compile, not tested on hardware.

This PR tries to make it possible for a user to say
```
odrv0.config.enable_uart = True
odrv0.config.move_uart_to_gpio_3_and_4 = True
odrv0.save_configuration()
odrv0.reboot()
```
After the reboot, the ODrive should talk UART on GPIO pins 3 and 4 instead of the default 1 and 2.

The intention is to enable this on v3.5 or later only, since earlier versions use GPIO 3 and 4 for step/dir.


samuelsadok:
Superseded by #492 which takes advantage of the recent low-level refactor.

PR #389, opened 2019-12-10
Add a Dockerfile which builds in an Ubuntu container

emdash:
I could not get the firmware to build correctly on Debian owing to some arcane include search path issues I would rather not spend time debugging. I decided to try the `Dockerfile` route rather than deal with the pain of setting up an entire VM for FW development, so that I can build using the embedded arm toolchain from the Ubuntu PPA.

You can build the firmware by invoking `docker build .` from the project root. For now, build artifacts need to be manually copied out of the resulting container.


emdash:
Going to re-open this on WetMelon's RazorsEdge branch.

PR #393, opened 2019-12-12
UART handler improvements

sgilbert182:
Added templated circular buffer code for quickly adding self contained circular buffers

Changed UART handler to use a wake/suspend mechanism and to use the DMA to transfer 2 bytes at a time. Using the half and full transfer complete interrupts, each character received can be read out of the UART register immediately, significantly reducing UART register overwrite conditions. The two DMA ISR's can then transfer the received characters into a circular buffer and signal the UART task should wake up and read the data back out of the circular buffer at it's earliest convenience. The task can wake, handle the received characters and then go back to sleep. This method reduces the data handling overhead of received characters and allows the task to sleep unless needed.

The transfer error handling has been moved in to the error detection ISR to allow for immediate handling of UART errors rather than waiting for the task to wake up and handle the UART.

Menu function has been split up into individual functions and the if else tree has been replaced with a switch statement allowing for faster code execution and easier code maintenance.


Wetmelon:
Thanks for  the PR, it looks awesome!  How much testing have you done with it?  Any noticeable performance benefits in speed or stability?  It's clearly a big benefit in readability and maintainability :)

ghost:
Hi, I'm glad it looks useful! 

The aim was to improve the responsiveness of the serial port (someone raised it as a potential issue). The task no longer has a longest time of 1ms delay + waiting for the receive buffer full interrupt between executions. It takes as long as the scheduler takes to get round to it after a single character has been received now. I had planned on doing a comparison using a scope and a raising gpio on the DMA handler and resetting it on task execution but may have to look at the actual uart rx pin rather than the DMA isr to get a better view on latancy. 

The case statement approach should give a better execution time under debug but will be similar when optimisation is turned on. More about readability there. 

As for stability, the DMA is pretty stable so there's probably not much of an improvement there. The 2 char wide buffer just improves how fast chars can be pulled in and handled. However the restart under error will be much faster as it doesn't have to wait until the task to restart the serial port. 

The circular buffer code has gone through a lot of testing and linting on previous projects. There is an issue with buffers not being x^2 and writing large blocks of data where counters could overflow (see comments in the init code) 

The uart suspend/wake method I've used a lot on previous projects and tested here by slinging back and fourth data using an echo function (not committed here) with no character drops. With so few tasks being executed here there should a significant improvement in latancy. 

The menu I tested by using break points in each function and typing in the relevant command to see if it got called correctly so far so good.

ghost:
I forgot to add, risk of data loss has been reduced as you need to be receiving the 3 character before the first one has been handled for a character to be dropped. At 2*(1/11520) s for 2 full chars that's a lot of processing time. By handling byte by byte, data loss should only affect the char in the buffer, the overwrite char and any chars sent before the uart is restarted. 
The time taken to transfer the data to the ring buffer is very short so can be run as a high priority ISR if need be.

sgilbert182:
Apologies for any confusion, the above posts are mine. I auto signed in from my phone and must have had a previous account once apon a time

madcowswe:
Hi,
Had a very quick look and thanks for putting in a lot of effort on this, looks clean and organized. Big improvement!

Firmware dev is on hold from me for a while while I focus on getting ODrive v4 out, so I have to keep this PR hanging for a bit. Sorry about that. I am looking forward to go through it at a later date, so please keep it open.

Thanks.

sgilbert182:
Hi,
Not a problem will leave it open. Sorry it's not in your usual code format, would you like me to correct that?
Cheers
Simon

samuelsadok:
Yes if you could adapt the formatting that would be nice, otherwise I can do it too. It seems to be mostly the curly braces.

Firmware/communication/interface_uart.cpp
```diff
@@ -25,88 +26,141 @@ osThreadId uart_thread;
...
+            }
         }
         return 0;
     }
 
     size_t get_free_space() { return SIZE_MAX; }
 private:
     uint8_t tx_buf_[UART_TX_BUFFER_SIZE];
+    uint8_t TXBuffer[UART_TX_BUFFER_SIZE];
+    CCBBuffer<uint8_t> CTXCircularBuffer;
```
samuelsadok, line 73: these seem to be unused

sgilbert182:
Sure, I'll take a look at it and see what I can do. I noticed you merged the latest develop in to this pull request too, that should make things easier!

Firmware/communication/ascii_protocol.cpp
```diff
@@ -87,192 +104,291 @@ void ASCII_protocol_process_line(const uint8_t* buffer, size_t len, StreamSink&
...
-            if (numscan >= 3)
-                axis->controller_.input_vel_ = vel_feed_forward;
-            if (numscan >= 4)
-                axis->controller_.input_current_ = current_feed_forward;
-            axis->controller_.input_pos_updated();
-            axis->watchdog_feed();
-        }
+    switch(cmd[0])
+    {
+        case 'p': setPosition(cmd, response_channel, use_checksum);             break;  // position control
```
samuelsadok, line 109: Unaligned breaks are preferred here, see https://github.com/madcowswe/ODrive/pull/410#discussion_r427725697

Firmware/communication/ascii_protocol.cpp
```diff
@@ -26,20 +26,36 @@
...
     output.process_bytes((uint8_t*)response, len, nullptr); // TODO: use process_all instead
-    if (include_checksum) {
+    if (include_checksum)
+    {
         uint8_t checksum = 0;
         for (size_t i = 0; i < len; ++i)
             checksum ^= response[i];
-        len = snprintf(response, sizeof(response), "*%u", checksum);
-        output.process_bytes((uint8_t*)response, len, nullptr);
+        output.process_bytes((uint8_t*)response, snprintf(response, sizeof(response), "*%u", checksum), nullptr);
```
samuelsadok, line 58: I'd say it's better not to hide complex and/or time/memory-consuming behavior such as `snprintf` inside an argument but instead keep it on a separate line

Firmware/communication/interface_uart.cpp
```diff
@@ -25,88 +26,141 @@ osThreadId uart_thread;
...
-                return -1;
-            buffer += chunk;
-            length -= chunk;
-            if (processed_bytes)
-                *processed_bytes += chunk;
+            }
+            else
+            {
+                // transmit chunk
+                if (HAL_UART_Transmit_DMA(&huart4, (uint8_t *)memcpy(tx_buf_, buffer, chunk), chunk) != HAL_OK)
```
samuelsadok, line 51: same here, better to keep `memcpy` on a separate line

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+    size_t usedSpaceLinear();
+    bool flushBuffer();
+    int32_t write(const dataType_t * const elem, size_t length);
+    int32_t read(dataType_t * elem, size_t length);
+    bool readNewest(dataType_t * pData);
+
+private:
+    circularBuffer_t m_cb;
+    bool m_useMutex;
+    bool m_overwriteOldData;
```
samuelsadok, line 90: Since `m_overwriteOldData` is known at compile time and `CCBBuffer` is already templated, you could make it a template parameter. Same with `m_cb.tracker.numElements`. `m_useMutex` is unused.

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+ * \param   overwriteOldData    - enables discarding of old data. Defaults to false
+ *
+ * \return  None
+ */
+template <class dataType_t>
+CCBBuffer<dataType_t>::CCBBuffer(dataType_t * pArray, const size_t numElements, const bool useMutex, const bool overwriteOldData)
+    : m_cb {{numElements, 0, 0}, pArray}
+    , m_overwriteOldData(overwriteOldData)
+{
+//        assert(this->isPowerOfTwo(size), "The size must be of ^2 otherwise overflow condition of start/end tracker values will corrupt data in buffer ");
```
samuelsadok, line 124: if `m_cb.tracker.numElements` is turned into a template parameter you could use `static_assert` if you were to keep the check. However it looks like it's obsolete.

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+ */
+template <class dataType_t>
+size_t CCBBuffer<dataType_t>::remainingSpaceLinear()
+{
+size_t length = 0;
+
+/* if end is less than start index then data goes from start pointer to size of
+ * array and start of array to end index. free space exists between end index
+ * and start index. If start is less than end index, then data goes from end
+ * index to numElements of array and start of array to start index.
```
samuelsadok, line 179: I think you meant "If start is less than **or equal to** end index, then **free space** goes from end index to numElements of array and start of array to start index." (or something equivalent)

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+    typedef struct
+    {
+        CBTracker_t tracker;                                                    /* buffer size and start/end counters */
+        dataType_t * pArray;                                                    /* pointer to array of elements. to be set on buffer init */
+    }circularBuffer_t;
+public:
+
+public:
+
+    CCBBuffer(dataType_t * pArray = nullptr, const size_t size = 0, const bool useMutex = false, const bool overwriteOldData = false);
```
samuelsadok, line 74: Providing default arguments here makes it look like `pArray = nullptr` is a valid use case, but its not (`flushBuffer()` -> `memset(nullptr, ...)` would cause undefined behavior)

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+*******************************************************************************/
+
+template <class dataType_t>
+class CCBBuffer
+{
+    typedef struct
+    {
+        size_t numElements;                                                     /* maximum number of elements */
+        int32_t start;                                                          /* index of oldest element */
+        int32_t end;                                                            /* index at which to write new element */
```
samuelsadok, line 62: This should most likely be `size_t` to avoid signed/unsigned conversion in various places. Same with `readCnt` and `writeCnt` further down and function return values.

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+{
+    int32_t readCnt = 0;
+    size_t toRead = 0;
+
+    if(m_useMutex)
+    {
+
+    }
+
+    for (readCnt = 0; (!isEmpty() && readCnt < (int32_t)length); readCnt =+ toRead)
```
samuelsadok, line 296: `readCnt =+ toRead` is the same as `readCnt = (+toRead)`. Probably you meant `+=`. As far as I can see in practice the loop will still terminate fine most of the time because `isEmpty()` will be true after the second iteration but in some cases it could cause a race condition.

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+ *          end OR string array which ever is reached first (no wrap around)
+ *
+ * \param   None
+ *
+ * \return  Used space in buffer
+ */
+template <class dataType_t>
+size_t CCBBuffer<dataType_t>::usedSpaceLinear()
+{
+    return (size_t)(((m_cb.tracker.end < m_cb.tracker.start) ? (int32_t)m_cb.tracker.numElements : m_cb.tracker.end) - m_cb.tracker.start);
```
samuelsadok, line 211: Possible race condition: assume `m_cb.tracker.end < m_cb.tracker.start` evaluates to false and then a UART RX interrupt happens and makes `m_cb.tracker.end` wrap around. After the interrupt this function will compute `m_cb.tracker.end - m_cb.tracker.start` but end is now smaller than start so this will cause an underflow.

Suggested fix: read `m_cb.tracker.end` into a local variable first and then only use the local variable.

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+DEFINITIONS
+*******************************************************************************/
+
+typedef enum
+{
+	CB_OK = 0,
+	CB_FAIL,
+	CB_SIZE_ERROR,
+	CB_POINTER_ERROR
+}CB_Status_t;
```
samuelsadok, line 41: this seems to be unused

Firmware/communication/CircularBuffer.hpp
```diff
@@ -0,0 +1,333 @@
...
+        size_t numElements;                                                     /* maximum number of elements */
+        int32_t start;                                                          /* index of oldest element */
+        int32_t end;                                                            /* index at which to write new element */
+    }CBTracker_t;
+
+    typedef struct
+    {
+        CBTracker_t tracker;                                                    /* buffer size and start/end counters */
+        dataType_t * pArray;                                                    /* pointer to array of elements. to be set on buffer init */
+    }circularBuffer_t;
```
samuelsadok, line 69: What is the reason for nesting these fields inside structs instead of making them top level member fields?

Firmware/communication/ascii_protocol.cpp
```diff
@@ -26,20 +26,36 @@
 
 /* Private variables ---------------------------------------------------------*/
 /* Private function prototypes -----------------------------------------------*/
+
+void setPosition(char * pStr, StreamSink& response_channel, bool use_checksum);
```
samuelsadok, line 30: Our naming convention for functions is snake_case (e.g. `set_position`). To make it clear that these functions belong together as a group you could add a prefix (`cmd_set_position` or similar).

samuelsadok (commented):
Alright here's a more detailed review.

I ran it against our [test suite](https://github.com/madcowswe/ODrive/blob/devel/tools/odrive/tests/uart_ascii_test.py) and it passes mostly, except [`TestBurnIn`](https://github.com/madcowswe/ODrive/blob/devel/tools/odrive/tests/uart_ascii_test.py#L182-L210) which failed 3 out of 50 times. `r vbus_voltage` returned `unknown command` or `invalid property` in these instances. Usually I don't run the test 50 times but seemingly by coincidence I ran into the issue on the first try. For reference, `devel` failed this test 0 out of 50 times. (You don't have to run the tests, they are not polished enough for general use).

There's a non-zero chance that one of the review comments addresses the issue.

Firmware/communication/CircularBuffer.hpp
sgilbert182: This was a hold over from when the status of each function would be returned but I have removed this.

Firmware/communication/CircularBuffer.hpp
sgilbert182: Start and End should be size_t. as should readCnt and writeCnt but the return values were deliberately made signed to be able to return -1 as an error indicator. Is it worth keeping leaving the function returns as int so the error indicators can be completed later?

Firmware/communication/CircularBuffer.hpp
sgilbert182: No real reason here but was a hold over from when it was C code and not templated. I will flatten the structure here.

Firmware/communication/CircularBuffer.hpp
sgilbert182: Again a hold over from an earlier project where this class had an init function. The design allowed for the class to be instantiated before the memory had been allocated, The memory could then be allocated and passed in later. Would you prefer an init function and a trap/check on null ptr in the ctr or would you prefer the removal of the default arguments?

Firmware/communication/CircularBuffer.hpp
sgilbert182: m_overwriteOldData is known for the current project and as it stands but templating it would mean it couldn't be changed during runtime. Also, templating it would cause duplication of code so it may become a speed/size optimisation issue.
m_cb.tracker.numElements was to be flexible to allow for changes in buffer size. Not needed here but has been used in a dynamic filter library. If however, you want this as a templated argument then the whole memory allocation could be internalised and the pointer and size allocation in the constructor could be removed.

Firmware/communication/CircularBuffer.hpp
sgilbert182: Obsolete but left in for compatibility of previous work. Static assert is tempting but relies on tempating parameters which I'm not fond of reasons described in previous comments.

Firmware/communication/CircularBuffer.hpp
sgilbert182: It's so good to be working with someone who reviews code properly! I will double check but you're probably right.

Firmware/communication/CircularBuffer.hpp
sgilbert182: Good spot, this hasn't been an issue due to the speed at which the buffer is read from vs how infrequently the ISR is triggered (relatively speaking). Ideally this needs the mutex protection to be finished off for anything faster than UART unless you have another recommendation on this.

Firmware/communication/ascii_protocol.cpp
sgilbert182: Force of habit, will change to snake case in the future and look at what prefix is appropriate. As these functions are local not global, would you still want a prefix?

Firmware/communication/ascii_protocol.cpp
sgilbert182: Force of habit, will adjust.

Firmware/communication/CircularBuffer.hpp
sgilbert182: Definitely a typo, will correct.

Firmware/communication/ascii_protocol.cpp
sgilbert182: I blame my OCD but happy to change it.

Firmware/communication/interface_uart.cpp
sgilbert182: Seems to be missing a couple of code tidy changes, I will take a look

Firmware/communication/CircularBuffer.hpp
samuelsadok: I think calculating with unsigned internally and leaving the return value signed is a legitimate solution. For this there's actually a type called `ssize_t`.
Personally I've come to favor error codes as return values and return the actual output as an output argument. That way, error codes can be properly typed (an enum) and the approach works in a broader range of cases (where the output value is something else than a length.
But here both are fine.

Firmware/communication/CircularBuffer.hpp
samuelsadok: Technically it's possible to have one templated implementation to cover both static and dynamic use cases. [Here's a quick test](https://godbolt.org/z/rrAzMG). In the static case we see that the compiler is indeed aware that `size_` is a power of two, as it resolves `divide_by_size` to a simple right shift. With this approach the allocation would still have to be external which I think is quite ok.
I'm not too concerned about the size as long as we're not using 10 different instantiations throughout the firmware.
In general it certainly makes sense for you to use the same implementation throughout your projects but in any case it's somewhat likely that they will diverge sooner or later.

Firmware/communication/CircularBuffer.hpp
samuelsadok: I think copying `m_cb.tracker.end` to a local variable before using it should do the job. Worst case you're reporting less `usedSpaceLinear` than the true value and as a result, `read()` will run for another iteration (or return less than all data). But everything is still well-behaved.

In general our thread-safety considerations are made simpler if we consider three things:
 - The producer only advances the end-pointer _after_ writing valid data into the circular buffer.
 - The consumer only advances the start-pointer _after_ copying data out of the circular buffer.
 - In our particular case the consumer can be interrupted by the producer but not the other way around. To be fair this assumption is a bit sketchy and should probably be written in a comment. But maybe this assumption is not even needed, didn't check that.

Firmware/communication/ascii_protocol.cpp
samuelsadok: Not necessarily, up to you. Oh btw since they are local you could prepend `static`.

Firmware/communication/ascii_protocol.cpp
madcowswe: In this case the break column improves readability I think, so lets keep it like that.

samuelsadok:
As you may have seen, we merged your ASCII protocol cleanup in a separate PR recently (#447).

As for the UART interrupt + cicular buffer architecture we arrived at the conclusion that having the UART RX interrupt fire for every incoming byte eats too much CPU (and is unpredictable).

A nice approach would be to use double-buffered DMA and use the RX-not-empty interrupt to unblock the UART RX thread. The RX-not-empty interrupt would be disabled in the interrupt handler and reenabled in the UART RX thread. This means that if multiple bytes arrive before the UART RX can process them the interrupt does not fire for every byte.

However, as an alternative, lower effort approach, we just increased the polling frequency to 8kHz (#449).

So I'm closing this PR now.

Thank you for your contribution on the ASCII cleanup!

PR #395, opened 2019-12-18
Add enable pin functionality

Akex2:
Add enable pin

### DIFF
--- a/Firmware/MotorControl/axis.cpp
+++ b/Firmware/MotorControl/axis.cpp
@@ -68,6 +68,7 @@ static void step_cb_wrapper(void* ctx) {
 // @brief Sets up all components of the axis,
 // such as gate driver and encoder hardware.
 void Axis::setup() {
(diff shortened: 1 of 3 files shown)


samuelsadok:
With this the ODrive would require a rising edge on "enable" _after_ the startup procedure in order to enter closed loop control. However that means whatever controls the enable pin now needs feedback from the ODrive to be able to trigger enable at the right time.
Shouldn't we rather immediately jump to closed loop control if someone hardwires the enable pin to high? If so, we could do continuous polling of the enable pin in `run_control_loop` instead of interrupts.

samuelsadok:
Checking the code again I see that you're actually not relying on interrupts but call `Axis::enable_pin_cb()` in every control iteration so my previous concern is invalid.

However we're brushing up the related PR #281 now so I'm gonna close this one. Feel free to check  out and comment on the other PR if you're interested.

Issue #231, opened 2018-09-01
Read UART on interrupt


Nate711:
Currently there's a worst-case delay of 1ms between receiving a uart message and actually reading and acting upon it. To fix this, the odrive could trigger an interrupt when it detects a new uart byte, or better, when it detects that a string of bytes have finished sending (idle line interrupt). 

The idle line interrupt can be set up like this:
```cpp
void start_uart_server() {
    __HAL_UART_ENABLE_IT(&huart4, UART_IT_IDLE);   // enable idle line interrupt
   // rest of code...
}
```

And then you can handle the interrupt like so:
```cpp
void UART4_IRQHandler(void)
  /* USER CODE BEGIN UART4_IRQn 0 */
  if(huart4.Instance->SR & UART_FLAG_IDLE) {
    // Trigger the uart server thread to read the dma 
  }

  /* USER CODE END UART4_IRQn 0 */
  // HAL_UART_IRQHandler(&huart4);
  /* USER CODE BEGIN UART4_IRQn 1 */

  /* USER CODE END UART4_IRQn 1 */
}
```

What I wasn't able to get to work was clearing the interrupt flags successfully. Inside the if statement above I had the code:
```cpp
volatile uint32_t tmp;                                   /* Must be volatile to prevent optimizations */
tmp = huart4.Instance->SR;                       /* Read status register */
tmp = huart4.Instance->DR;
if(uart_thread) {
   osSignalSet(uart_thread, 0x01);
}
```
But if I remember correctly the flags were not being cleared successfully.

I also set up the uart server thread to wait on a thread signal like so:
```cpp
evt = osSignalWait(0x01, osWaitForever); // block until interrupt sets the flag
if (evt.status == osEventSignal) { // only query dma if got an event signal 
}
```

sgilbert182:
Hi, I found with the UART on the STM32's it was very effective to have a 2 byte wide circular receive buffer tied using DMA to the UART. Then have the DMA interrupt on every half and full receive. In that interrupt you can place the received data in your character queue. This gives you a nice double buffering of the UART and no need to wait for the idle interrupt plus as the DMA clears the UART receive buffer, you have a lot longer to process the incoming data.

sgilbert182:
It's also worth noting, the UART processing task has a periodic wake up of 1ms and is suspended in the meantime. That could be changed to a sleep/suspend and wake on UART interrupt giving much faster response.

madcowswe:
[...]

Issue #159, opened 2018-04-18
Watchdog feeding thread


madcowswe:
Configurable check for various timeouts, and takes safe actions.
If this thread is hung, the watchdog resets the CPU.

Wetmelon:
Fixed in #316

Wetmelon:
Did #316 take care of this, or is this something we still need to do?

sgilbert182:
Hi,
Has any work been done on the watchdog thread? If not, I have something that may be suitable.
Cheers
Simon

PR #400, opened 2020-03-09
Watchdog task

sgilbert182:
#231 #393 #159 

Added C++ watchdog aware RTOS Task base class.

This implementation offers a series of sources and headers that boil down to a single inheritable class that allows a user to set up a watchdog protected thread. If a watchdog is implemented and passed in as an argument along with a period to the child class, the back end automatically registers, deregisters, feeds and notifies 'task suspend state' to the watchdog thread. This allows the watchdog to monitor asynchronous as well as periodic threads/tasks.

funcMain() is the main task loop and must be implemented in the child class. No for or while loop needed as this is handled in the parent class.

funcBegin() and funcEnd() are override-able one shot functions called before and after funcMain respectively and may be used to initialise and close down peripherals and variables but are not required.

funcTimeout() is an override-able oneshot function that is automatically called when the task has failed a watchdog check and can be used to trigger a 'safe-ing action' before a reset (if returning false) or a task reset with no hardware reset (if returning true).

UARTControlTask takes the uart DMA solution previously submitted and re-implements it using the task template.

WatchdogTask is the implementation of the task monitoring task that checks each registered task for activity (if not sleeping). If the task appears to have locked up, it triggers a registered callback and then (if the callback returns false) allows the IWDG to starve and reset the system.

Moved circular buffer to a Utilities folder
Added an implementation of a linked list

### DIFF
--- a/Firmware/Board/v3/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_hal_iwdg.h
+++ b/Firmware/Board/v3/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_hal_iwdg.h
@@ -1,0 +1,237 @@
+/**
+  ******************************************************************************
+  * @file    stm32f4xx_hal_iwdg.h
+  * @author  MCD Application Team
+  * @brief   Header file of IWDG HAL module.
+  ******************************************************************************
+  * @attention
+  *
+  * <h2><center>&copy; COPYRIGHT(c) 2017 STMicroelectronics</center></h2>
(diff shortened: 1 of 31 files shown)


PAJohnson:
@sgilbert182 This PR is now outdated compared to `devel`. The merge of `control_loop_refactor` brought significant changes. Can you bring this PR up to date? The UART comms interface was also changed recently to run at the control loop rate of 8kHz instead of 1kHz - not sure if this changes your use case for the UART-related parts of this PR.

PR #401, opened 2020-03-09
Gpio debounce

sgilbert182:
Created task to debounce usbscribed GPIOs. Based on code recommendation found here: http://www.ganssle.com/debouncing-pt2.htm

Rewrote EXTI subscribe tools found in GPIO.c, converted to a C++ Class base class, EXTI child class, and Debounce child class. Then moved to their own source and header files.

Base class uses a linked list so GPIO's can properly be added and removed. Current method doesn't seem to completely remove pin details on unsubscribe causing a limited number of times the subscribe GPIO feature can be used between resets.

GPIO debounce task can subscribe a number of GPIO's to be debounced over rolling configurable window (currently 10ms), each sample taken 1ms apart. Each pin can have it's own callback triggered the moment the pin stabilises in it's asserted state. The callback is called once and won't be called again until the pin is de-asserted first. Suitable for endstop or emergency stop button detection.


sgilbert182:
Updated build list to specifically include stm32f405xx.h location.

sgilbert182:
Removed #include stm32f405xx.h from #include list for subscribebase.cpp

PAJohnson:
@sgilbert182 This PR has fallen behind `devel`. Can you bring it up to date?

sgilbert182:
Hi,

I'll take a look at updating this but it may not be until the weekend

Cheers
Simon

PR #402, opened 2020-03-29
Update encoders.md with AS5047/AS5048 interfaces.

BenChung:
The current documentation claims that the AS5048 supports an ABI interface, but there is no documentation for the device that substantiates that or provides use information. This clarifies that, specifying that only the AS5047 can be used with as an ABI encoder.


samuelsadok:
Since SPI encoder support is now implemented (on the `devel` branch) I replaced the note on AS5047/AS5048 by a section on SPI encoders. See [here](https://github.com/madcowswe/ODrive/blob/devel/docs/encoders.md#spi-encoders). So I'm gonna go ahead and close this.

PR #405, opened 2020-04-26
Change behaviour of get_inverter_temp

### LINKED ISSUE #404: get_inverter_temp through ASCII no response
Hi,

I use "r axis0.motor.get_inverter_temp" command for getting axis0 inverter temperature, but there is no response. Is it my command wrong? any suggestions? 
vbus_voltage and other commands are usable.

Thanks

Wetmelon:
* Save inverter temperature to a member variable
* Use the member variable to update thermal limits
* Change the communication endpoint `get_inverter_temp()` into a ro_property that returns the member variable


Firmware/MotorControl/motor.cpp
```diff
@@ -150,8 +150,7 @@ float Motor::get_inverter_temp() {
     return horner_fma(normalized_voltage, thermistor_poly_coeffs, thermistor_num_coeffs);
 }
 
-bool Motor::update_thermal_limits() {
-    float fet_temp = get_inverter_temp();
+bool Motor::update_thermal_limits(float fet_temp) {
```
samuelsadok, line 153: why was `float fet_temp = get_inverter_temp();` moved out of here instead of doing `inverter_temp_ = fet_temp;` here? Not that it makes a big difference.

Wetmelon: I'm not a big fan of putting HWIO operations inside logical functions.  IMO it's better architecture to have a clear delineation:

- Read Inputs
- Update Logic
- Write Outputs

PR #406, opened 2020-04-26
Documentation Updates

camrbuss:
Working off the Change-log to apply documentation updates


docs/commands.md
```diff
@@ -58,6 +65,7 @@ Possible values are:
 * `CTRL_MODE_POSITION_CONTROL`
 * `CTRL_MODE_VELOCITY_CONTROL`
 * `CTRL_MODE_CURRENT_CONTROL`
+* `CTRL_MODE_TRAJECTORY_CONTROL`
```
Wetmelon: There is no such control mode

camrbuss: Absolutely agree, that came from when I was reworking the documentation on `master`. Fixed with `573ad8f`

PR #410, opened 2020-05-14
RazorsFrozenTesting

samuelsadok:


Firmware/MotorControl/axis.hpp
```diff
@@ -60,11 +65,19 @@ class Axis {
...
                                     //   For M0 this has no effect if enable_uart is true
+        bool step_dir_always_on = false; //<! Keep step/dir enabled while the motor is disabled.
+                                         //<! This is ignored if enable_step_dir is false.
+                                         //<! This setting only takes effect on a state transition
+                                         //<! into idle or out of closed loop control.
+
         float counts_per_step = 2.0f;
 
-        float watchdog_timeout = 0.0f; // [s] (0 disables watchdog)
+        float watchdog_timeout = 0.0f;  // [s] (0 disables watchdog)
```
madcowswe: 0 disables is no longer true with enable_watchdog, right?

Firmware/MotorControl/encoder.cpp
```diff
@@ -163,7 +182,7 @@ bool Encoder::run_direction_find() {
 // TODO: Do the scan with current, not voltage!
 bool Encoder::run_offset_calibration() {
     static const float start_lock_duration = 1.0f;
-    static const int num_steps = (int)(config_.calib_scan_distance / config_.calib_scan_omega * (float)current_meas_hz);
+    static const int num_steps             = (int)(config_.calib_scan_distance / config_.calib_scan_omega * static_cast<float>(current_meas_hz));
```
madcowswe: the space before the "=" hurts me.

Firmware/MotorControl/encoder.cpp
```diff
@@ -199,8 +218,8 @@ bool Encoder::run_offset_calibration() {
 
     // scan forward
     i = 0;
-    axis_->run_control_loop([&](){
-        float phase = wrap_pm_pi(config_.calib_scan_distance * (float)i / (float)num_steps - config_.calib_scan_distance / 2.0f);
+    axis_->run_control_loop([&]() {
+        float phase   = wrap_pm_pi(config_.calib_scan_distance * static_cast<float>(i) / static_cast<float>(num_steps) - config_.calib_scan_distance / 2.0f);
```
madcowswe: This static_cast stuff makes it really hard to read, what's wrong with regular cast?

Firmware/MotorControl/encoder.cpp
```diff
@@ -276,7 +294,7 @@ static bool decode_hall(uint8_t hall_state, int32_t* hall_cnt) {
 }
 
 void Encoder::sample_now() {
-    switch (config_.mode) {
+    switch (mode_) {
```
madcowswe, line 297: why is there now a copy of `config.mode_`? I only see it written to in `setup()`. Is it intentional that you can only change encoder mode with a reboot?

Firmware/MotorControl/encoder.cpp
```diff
@@ -383,10 +543,12 @@ bool Encoder::update() {
 
     //// compute electrical phase
     //TODO avoid recomputing elec_rad_per_enc every time
-    float elec_rad_per_enc = axis_->motor_.config_.pole_pairs * 2 * M_PI * (1.0f / (float)(config_.cpr));
-    float ph = elec_rad_per_enc * (interpolated_enc - config_.offset_float);
+    float elec_rad_per_enc = axis_->motor_.config_.pole_pairs * 2 * M_PI * (1.0f / static_cast<float>(config_.cpr));
+    float ph               = elec_rad_per_enc * (interpolated_enc - config_.offset_float);
```
madcowswe: These spaces again..

Firmware/MotorControl/motor.hpp
```diff
@@ -187,6 +193,7 @@ class Motor {
     DRV8301_FaultType_e drv_fault_ = DRV8301_FaultType_NoFault;
     DRV_SPI_8301_Vars_t gate_driver_regs_; //Local view of DRV registers (initialized by DRV8301_setup)
     float thermal_current_lim_ = 10.0f;  //[A]
+    float inverter_temp_ = 20.0f;
```
madcowswe: I would say 0.0f would be a better default. Looks more "uninitialised" than 20.

Firmware/MotorControl/utils.hpp
```diff
@@ -89,6 +77,20 @@ static inline float fmodf_pos(float x, float y) {
...
+/**
+ * @brief Similar to modulo operator, except that the output range is centered
+ * around zero.
+ * The returned value is always in the range [-pm_range, pm_range).
+ */
+static inline float wrap_pm(float x, float pm_range) {
+    return fmodf_pos(x + pm_range, 2.0f * pm_range) - pm_range;
+}
+
+//beware of inserting large angles!
```
madcowswe: beware of large angles no longer applies

madcowswe (approved):
@samuelsadok After you have addressed the comments, feel free to merge to `devel` yourself when you are done.

Firmware/MotorControl/encoder.cpp
madcowswe: More instances of this exist throughout.

Firmware/MotorControl/encoder.cpp
madcowswe: Btw feel free to leave it if there is a very good reason (readability is high on my priority list)
More instances of this exist throughout.

Firmware/MotorControl/axis.hpp
samuelsadok: there's still a separate check for `watchdog_timeout == 0.0f`. I'll remove though it since it's redundant now.

Firmware/MotorControl/encoder.cpp
samuelsadok: Here are some interesting perspectives on this:
https://stackoverflow.com/questions/101958/code-formatting-is-lining-up-similar-lines-ok

After reading this I also gravitate to the no-alignment camp, perhaps with exceptions such as:
```
float v_alpha = c*v_d - s*v_q;
float v_beta  = c*v_q + s*v_d;
```
which is essentially a matrix multiplication on two lines.

I'm gonna do an unalign-pass on the most prominent files.

Firmware/MotorControl/encoder.cpp
samuelsadok: C++ style casts are more explicit in the sense that a C style cast can resolve to `static_cast`, `reinterpret_cast` or `const_cast`, depending on the involved types. Another argument is to make casts in general less subtle:
> An ugly operation should have an ugly syntactic form.
>  [— Bjarne Stroustrup](http://www.stroustrup.com/bs_faq2.html#static-cast)

And ugly they are. I think a sensible rule would be to use classic casts for numerical code and C++ style casts in all other cases.

Firmware/MotorControl/encoder.cpp
samuelsadok: Yes, SPI encoders need a special setup procedure which is only done at startup. Apart from that the encoder class currently contains way too much state and code paths for a runtime-switch to feel comfortable, for instance looking at the `pre_calibrated` logic.

Firmware/MotorControl/motor.hpp
samuelsadok: or NaN? Could surprise one or the other user but seems to me like the cleanest default.

Firmware/MotorControl/encoder.cpp
Wetmelon: fwiw that article was from 2008 and clang-format on git hook is a thing, so git merge isn't really an issue anymore, it's mostly personal preference now.

PR #418, opened 2020-06-03
Autogenerate Protocol Code

samuelsadok:
Previously, the variables and functions in the firmware were exposed on Fibre (the USB protocol) by complex template machinery. This PR introduces a YAML file which serves as the new ground truth for the API definitions and replaces the template machinery with autogenerated code based on the YAML file.

This brings several benefits:
 - Drastically faster connect times (20s -> 2s). This is because the JSON that is fetched by odrivetool now lies in the flash as a flat array instead of being "calculated" at runtime for every USB request.
 - Paves the way for clean API documentation that can be turned into autogenerated HTML webpages.
 - Gives users a clean possibility for autogenerating static library code for host-side software (the current approach used by odrivetool does not work well with ahead-of-time compiled languages).
 - Foundation for a future low-level protocol revision (function-oriented, dynamic object tree)

The protocol itself is not changed, the new firmware is backwards compatible with old versions of odrivetool.


madcowswe (approved):
Looks good to me.
Just make sure:
- [x] Decent test coverage of the various protocol features 
  - [x] Read properties
  - [x] Write properties
  - [x] Write hooks
  - [x] Function calls
  - [x] p/v/c uart commands
  - [x] r/w uart commands
- [x] New dependencies/requirements for compiling is communicated well to developers. This could be good error messages directing how to add new deps, or otherwise.

samuelsadok:
Ok I added a HWIL test specifically for the Fibre protocol and also added a write-hook test for the ASCII protocol.

If a developer is missing a python dependency, the error message looks like this now:
```
* 19) python interface_generator_stub.py --definitions odrive-interface.yaml --template fibre/cpp/function_stubs_template.j2 --output autogen/function_stubs.hpp                                                         
No module named 'jinja2'
Note that there are new compile-time dependencies since around v0.5.1.
Check out https://github.com/madcowswe/ODrive/blob/devel/docs/developer-guide.md#prerequisites for details.
 *** tup messages ***
 *** Command ID=33178 failed with return value 1
```

PR #419, opened 2020-06-05
implementation of motor thermistor support.

rowanG077:
I have started the implementation of allowing users to connect a thermistor that is attached to a motor. The basic steps I took was create a new abstract class which allows current limiting. And based on that create a thermistor current limiter. The FET thermistor has been moved over to this class as well.

I have added some functions to get the adc_channel given a GPIO port/pin. User can choose which pins to use for the motor thermistor as long as it's a valid adc_channel. To differentiate the internal fet thermistor configuration contains a flag to indicate that the fet thermistor are on board and thus the pin assignment is immutable.

todo:

- [x] Create function in odrivetool to calculate and set thermistor coefficients based on thermistor constants.
- [x] Test in hardware


Firmware/MotorControl/current_limiter.hpp
```diff
@@ -0,0 +1,89 @@
...
+class CurrentLimiter {
+public:
+    virtual ~CurrentLimiter() = default;
+    virtual float apply_current_limit(float current_limit) = 0;
+};
+
+class ThermistorCurrentLimiter : public CurrentLimiter {
+public:
+    struct Config_t {
+        std::array<float, 4> thermistor_poly_coeffs = { };
```
samuelsadok: Feels a bit scary to store an STL class to NVM. We don't have a guarantee that it doesn't contain any funny run-time variables. I'd prefer a classical C array.

Firmware/MotorControl/current_limiter.hpp
```diff
@@ -0,0 +1,89 @@
...
+            make_protocol_property("error", &error_),
+            make_protocol_function("get_temp", *this, &ThermistorCurrentLimiter::get_temp),
+            make_protocol_function("set_pin", *this, &ThermistorCurrentLimiter::set_pin, "GPIO pin number to use for thermistor."),
+            make_protocol_ro_property("gpio_pin", &gpio_pin_),
+            make_protocol_object("poly_coefficients",
+                make_protocol_property("idx_0", &config_.thermistor_poly_coeffs[0]),
+                make_protocol_property("idx_1", &config_.thermistor_poly_coeffs[1]),
+                make_protocol_property("idx_2", &config_.thermistor_poly_coeffs[2]),
+                make_protocol_property("idx_3", &config_.thermistor_poly_coeffs[3])
+            ),
```
samuelsadok: For `odrivetool` to properly back up the configuration, all NVM-stored config variables need to be exposed as `[...].config.[...]`.

Also as a heads-up: With #418, either `thermistor_poly_coeffs` would need to be turned into a class (like [`timing_log_`](https://github.com/madcowswe/ODrive/blob/feature/interface_autogen/Firmware/MotorControl/motor.hpp#L123-L127)) or the variables would have to be named `poly_coefficients_idxN` instead of `poly_coefficients.idxN` (although the latter naming is cleaner). Just saying this so we can avoid API breakage when #418 gets merged.

Firmware/MotorControl/current_limiter.hpp
```diff
@@ -0,0 +1,89 @@
...
+        config->enabled = true;
+        config->onboard = true;
+    }
+
+    // Communication protocol definitions
+    auto make_protocol_definitions() {
+        return make_protocol_member_list(
+            make_protocol_property("error", &error_),
+            make_protocol_function("get_temp", *this, &ThermistorCurrentLimiter::get_temp),
+            make_protocol_function("set_pin", *this, &ThermistorCurrentLimiter::set_pin, "GPIO pin number to use for thermistor."),
```
samuelsadok: This should probably be a setter hook function of the `config.gpio_pin` property, which should be read-write. Example: [`abs_spi_cs_gpio_pin`](https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/encoder.hpp#L150-L151).
(btw the string is intended to act as argument name rather than an argument description)

Firmware/MotorControl/motor.hpp
```diff
@@ -185,7 +186,8 @@ class Motor {
             make_protocol_property("DC_calib_phC", &DC_calib_.phC),
             make_protocol_property("phase_current_rev_gain", &phase_current_rev_gain_),
             make_protocol_ro_property("thermal_current_lim", &thermal_current_lim_),
-            make_protocol_function("get_inverter_temp", *this, &Motor::get_inverter_temp),
+            make_protocol_object("fet_thermistor", fet_thermistor_.make_protocol_definitions()),
```
samuelsadok: We should add a new section "API migration notes" to CHANGELOG.md to keep track of these changes.

tools/odrive/utils.py
```diff
@@ -31,6 +32,25 @@
 class OperationAbortedException(Exception):
     pass
 
+def set_motor_thermistor_coeffs(motor, Rload, R_25, T_25, Beta, Tmin, TMax):
```
samuelsadok: Maybe add a brief note somewhere in the `docs/` that describes how to set up the thermistor.

Firmware/MotorControl/low_level.cpp
```diff
@@ -63,7 +63,7 @@ static uint16_t GPIO_port_samples [2][num_GPIO];
 *   Brake resistor PWM:
 *     Timer2.CCR3 (counter compare register 3)
 *     Timer2.CCR4 (counter compare register 4)
-* 
+*
```
samuelsadok: There are a couple of whitespace-only changes. While they are generally sensible we should do this in a separate formatting PR (and also enforce it by CI) to avoid noise here.

Firmware/MotorControl/current_limiter.cpp
```diff
@@ -0,0 +1,84 @@
...
+ThermistorCurrentLimiter::ThermistorCurrentLimiter(const Config_t& config) :
+        config_(config),
+        error_(ERROR_NONE)
+{
+    // For onboard thermistors the gpio_pin is the direct ADC channel
+    // because it's connected to a GPIO that's not exposed to the user.
+    // We set the gpio_pin_ to UINT16_MAX so it's clear that this
+    // value is not a "real" gpio pin.
+    if (config_.onboard)
+    {
```
samuelsadok: Our convention is to keep braces on the same line. See [here](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions) and [here](https://google.github.io/styleguide/cppguide.html#Conditionals).

samuelsadok (commented):
Thanks for the PR, looks promising! I added a couple of inline comments and here are two more general comments:

1. There are some conceptual differences between onboard and off board thermistors. The former is an implementation detail and doesn't need to be exposed on the API. But if so, it should likely be immutable. As such I think it makes sense to split either `ThermistorCurrentLimiter` or `ThermistorCurrentLimiter::Config_t` into two classes, one that samples and exposes a temperature and one that calculates current limit based on that. The `motor_thermistor`'s config would be exposed on the API but not the `fet_thermistor`'s config.


2. It would be good to be able to query `ThermistorCurrentLimiter` multiple times per control cycle and get consistent results. This could be done by having a separate sampling function that stores the temperature in a member variable. So the interface I'd suggest would be:

       class ThermistorCurrentLimiter {
           void update(); // updates the `temp_` member variable based on ADC. Called from `Axis::do_updates()`. Not needed if latching is already done in a Thermistor class.
           bool do_checks(); // updates the `error_` member variable based on `temp_`. Called from `Axis::do_checks()`. Can also be assimilated into `update()`.
           float get_current_limit(float user_limit); // pure function, no member variables are changed.
       };

Firmware/MotorControl/current_limiter.cpp
rowanG077: Will fix.

Firmware/MotorControl/current_limiter.hpp
rowanG077: According to cppreference it's guaranteed that `std::array` only non-static data member is semantically a plain C array.
https://en.cppreference.com/w/cpp/container/array

But your point stands. It's more obviously correct to just have a C-style array here.

Firmware/MotorControl/current_limiter.hpp
rowanG077: Will fix.

Firmware/MotorControl/current_limiter.hpp
rowanG077: Will fix. I will rename the variables.

Firmware/MotorControl/low_level.cpp
rowanG077: Yes it was done automatically by my editor. I will pull these changes out.

Firmware/MotorControl/motor.hpp
rowanG077: Will do.

tools/odrive/utils.py
rowanG077: Will do.

rowanG077:
@samuelsadok Thank you for your review! I will fix the points you mentioned.

1. This is true. I thought about it but the return type of `make_protocol_definition` is inferred. I guess because the actual type is insanely complex due to template magics. But this makes it hard to split into two classes because then I would need an abstract `Thermistor` class which does have an abstract method for `make_protocol_definition`. But you can't have auto type for abstract methods. Besides that I don't think the types for different protocol definitions would match.  

2. What do you mean exactly by `Not needed if latching is already done in a Thermistor class.` in your proposed API?

tools/odrive/utils.py
```diff
@@ -31,6 +32,25 @@
...
+    V = Rload / (Rload + R_temps)
+
+    fit = np.polyfit(V, temps, 3)
+    p1 = np.poly1d(fit)
+    fit_temps = p1(V)
+
+    motor.motor_thermistor.poly_coefficients.idx_0 = float(fit_temps[0])
+    motor.motor_thermistor.poly_coefficients.idx_1 = float(fit_temps[1])
+    motor.motor_thermistor.poly_coefficients.idx_2 = float(fit_temps[2])
+    motor.motor_thermistor.poly_coefficients.idx_3 = float(fit_temps[3])
```
madcowswe: I think the polynomial coefficients are in `fit` or `p1`, not in fit_temps.

madcowswe: Either way, to avoid code duplication, we should have thermistors.py call this function (or other way around)

Firmware/MotorControl/current_limiter.hpp
samuelsadok: Ah interesting, I didn't know that

samuelsadok:
@rowanG077

1. yeah this appears to be not quite straight forward but it's possible. You're right that return type of `make_protocol_definition` is very complex. However this function doesn't have to be declared in the base class, it can be declared and defined just in the deriving, non-abstract classes. 

   [Here's a draft](https://gist.github.com/samuelsadok/f1dbc3e2598d56127a6ff1464e2a2019) of an approach that I think would work. It doesn't introduce a Thermistor class but rather two child classes of `ThermistorCurrentLimiter`. Not 100% sure if it's an improvement but it might be. It leads to some code duplication but on the other hand it allows removing other code (constructors bodies, `set_pin`, `load_fet_default_thermistor_config`).

2. If there is a `Thermistor` class  with an update/sample function that stores ("latches") the measurement into a `Thermistor::temp_` variable, then the `ThermistorCurrentLimiter` class doesn't need to store the temperature itself.

rowanG077:
@samuelsadok 

1. I ended up creating an abstract `ThermistorCurrentLimiter` get's some pointers/references to configuration values and then creating the onboard and offboard classes like you suggested. This way there is no code duplication. I also moved the thermistors to axis and added a specific array for current limiters and thermal sensors. The `motor` class now only queries the current limiters to set the minimum current limit. This way there is a a nice path forward for other current limiters.

2. Clear.

I have added a changelog entry and created some docs. I'm not sure how I can build the docs so I have not tested whether the docs are working as intended.

I have now glued my thermistors to my motors but I'm waiting for the PCB I made to connect it to the ODrive. Once this is done I will test it in hardware.

Please let me know any other comments you have.

rowanG077:
Oh and another point. Should this be merged to the devel branch instead of master? I just now read about the devel branch on the forums.

Firmware/MotorControl/axis.cpp
```diff
@@ -20,13 +30,23 @@ Axis::Axis(int axis_num,
...
       sensorless_estimator_(sensorless_estimator),
       controller_(controller),
+      fet_thermistor_(fet_thermistor),
+      motor_thermistor_(motor_thermistor),
       motor_(motor),
-      trap_(trap)
+      trap_(trap),
+      current_limiters_(make_array(
+          static_cast<CurrentLimiter*>(&fet_thermistor),
+          static_cast<CurrentLimiter*>(&motor_thermistor))),
```
samuelsadok, line 37: as far as I can see you can do the same without `make_array`:
`current_limiters_{static_cast<CurrentLimiter*>(&fet_thermistor), static_cast<CurrentLimiter*>(&motor_thermistor)}`

Firmware/MotorControl/thermistor.hpp
```diff
@@ -0,0 +1,132 @@
...
+        ERROR_NONE = 0x0000,
+        ERROR_OVER_TEMP = 0x0001,
+    };
+
+    ThermistorCurrentLimiter() = delete;
+    virtual ~ThermistorCurrentLimiter() = default;
+    ThermistorCurrentLimiter(const ThermistorCurrentLimiter& other) = delete;
+    ThermistorCurrentLimiter(ThermistorCurrentLimiter&& other) noexcept = delete;
+    ThermistorCurrentLimiter& operator=(const ThermistorCurrentLimiter& original) = delete;
+    ThermistorCurrentLimiter& operator=(ThermistorCurrentLimiter&& original) = delete;
```
samuelsadok: IMO for brevity we can omit these declarations. I mean it's true that it's not allowed to copy/move a `ThermistorCurrentLimiter` object but in other classes (Motor, Axis, ...) we also don't state this explicitly. Btw I think copy/move assignment operators are not autogenerated for classes that contain references (because the reference itself is const).

Firmware/MotorControl/thermistor.hpp
```diff
@@ -0,0 +1,132 @@
...
+class OffboardThermistorCurrentLimiter : public ThermistorCurrentLimiter {
+public:
+    static const size_t num_coeffs_ = 4;
+
+    struct Config_t {
+        float thermistor_poly_coeffs[num_coeffs_] = { };
+        uint16_t gpio_pin = 4;
+        float temp_limit_lower = 100;
+        float temp_limit_upper = 120;
+        bool enabled = false;
```
samuelsadok, line 58: Technically a thermistor can just be disabled by putting the limits to infinity. On the other hand there is some merit to having an explicit enable flag. In any case we should make it consistent between onboard and offboard thermistors.

Firmware/MotorControl/thermistor.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void ThermistorCurrentLimiter::update() {
+    const float voltage = get_adc_voltage_channel(adc_channel_);
+    const float normalized_voltage = voltage / adc_full_scale;
+    temp_ = horner_fma(normalized_voltage, coefficients_, num_coeffs_);
+}
+
+bool ThermistorCurrentLimiter::do_checks() {
+    if (enabled_ && temp_ >= temp_limit_upper_) {
```
samuelsadok: We should keep the 5°C margin here that was in the old code to account for measurement noise.

Firmware/MotorControl/thermistor.cpp
```diff
@@ -0,0 +1,86 @@
...
+    num_coeffs_(num_coeffs),
+    temp_(),
+    temp_limit_lower_(temp_limit_lower),
+    temp_limit_upper_(temp_limit_upper),
+    enabled_(enabled),
+    error_(ERROR_NONE)
+{
+}
+
+float ThermistorCurrentLimiter::get_temp() const {
```
samuelsadok: I think this is not used anymore

samuelsadok (commented):
Looks good, just a few minor comments.

If you want to build the HTML docs you can check out [this](https://docs.odriverobotics.com/developer-guide#documentation). But for most purposes a simple markdown previewer will do (I'm using the VSCode extension "Markdown All in One" by Yu Zhang).

Yes PRs should go to `devel`, sorry I didn't notice.

Firmware/MotorControl/thermistor.hpp
rowanG077: I didn't include an enable for onboard thermistors because I didn't think we should support disabling them. An enable flag is definitely needed for the offboard thermistors because else we possibly sample unconnected IO and create bogus temps that then trigger the temperature limits.

docs/thermistors.md
```diff
@@ -0,0 +1,35 @@
+# Thermistors
```
samuelsadok, line 1: You can add this file to "docs/_data/index.yaml" so it will show up in the navigation bar.

Firmware/MotorControl/axis.cpp
rowanG077: @samuelsadok Yes you can but you lose some safety. For instance the following will compile without any warnings:

```
class Foo {
    Foo() : content({"test", "test2"}) {}
  
    std::array<std::string, 3> content;
};
```

The `make_array` helper ensures that you are forced to fill in every element else a compilation error is given. If you prefer it with a normal initializer_list I will change it.

Firmware/MotorControl/thermistor.cpp
rowanG077: Fixed.

docs/thermistors.md
rowanG077: Fixed

Firmware/MotorControl/thermistor.hpp
rowanG077: Fixed

Firmware/MotorControl/thermistor.hpp
rowanG077: Fixed. Added enable to onboard thermistors as well.

Firmware/MotorControl/thermistor.cpp
rowanG077: Fixed

Firmware/MotorControl/axis.cpp
samuelsadok: Ah I see. I assume in this example the default constructor would be used for the third string. Then perhaps in our case it makes sense to check for `nullptr` in the loops that iterate over these arrays. Either that or `make_array` can be moved to "utils.hpp".

rowanG077:
@samuelsadok 

I have implemented your suggestions and rebased on devel. Pull request target is now also devel. Next week on friday I will run hardware tests. Once this succeeds I will ping you so it can be merged.

rowanG077:
@samuelsadok 

I had some EMI issues with my motor which I was finally able to resolve by braiding the motor cables and adding a toroid. Anyway I tested the thermistor support in hardware today and it works perfectly. I tested the following scenarios succesfully:

1. Using new built-in helper function `set_motor_thermistor_coeffs` to calculate and set thermistor coefficients based on voltage divider circuit and thermistor spec.
2. Plotting temp for both fet and motor thermistors and they match what is expected.
3. Setting lower and upper temp limit low and ensure derating is functioning as expected.
4. Disabling/enabling a thermistor to check no overtemp is triggered and no current limiting is then applied due to that thermistor
5. Changing GPIO pin of motor thermistor.
6. Setting limits to pretty low values to check overtemp error is triggered.

If I'm missing some scenario please let me know so I can test that as well. I see that current unit has been removed from users in favor of specifying torque. Including the fet thermistor. Should I rewrite these PR to not be current limiting but torque limiting? I feel like current is better in this case since that is what is directly generating the heat and not the torque.

Firmware/odrive-interface.yaml
```diff
@@ -211,6 +211,8 @@ interfaces:
           WatchdogTimerExpired:
           MinEndstopPressed:
           MaxEndstopPressed:
+          OverTemp:
+            doc: Check `fet_thermistor.error` and `motor_thermistor.error` for more information.
```
samuelsadok: let's move this after `HomingWithoutEndstop` to reduce the likelihood of breaking someone's code silently

Firmware/odrive-interface.yaml
```diff
@@ -337,7 +380,6 @@ interfaces:
           BrakeDeadtimeViolation:
           UnexpectedTimerCallback:
           CurrentSenseSaturation:
-          InverterOverTemp:
           CurrentLimitViolation:
```
samuelsadok: again to reduce the likelihood of silent errors you can "pin" the bit number of the succeeding flag: `CurrentLimitViolation: {bit: 12}`

Firmware/odrive-interface.yaml
```diff
@@ -656,6 +694,12 @@ valuetypes:
       Homing:
         doc: run axis homing function
 
+  ODrive.ThermistorCurrentLimiter.Error:
+    values:
```
samuelsadok: For this enum the `flags` key should be used instead of `values` to make the interface generator chose numbers that are powers of two (see for example `ODrive.Motor.Error`). The `nullflag` key can then be used to specify the name of the "no flag is set"-state.

samuelsadok (commented):
Cool thanks!

Yes current limits make more sense when talking about thermal clamping. The `current_lim` is still exposed to the user and internally `effective_current_lim()` is still being honored so as far as I can see the new torque limit doesn't require any changes to this PR.

I added some minor comments on the YAML changes but apart from that it looks good to me. Regarding testing I think you covered all important aspects, so no further requests from my side.
I think @madcowswe also wants to take a look so give him a few days.

Firmware/odrive-interface.yaml
rowanG077: Fixed.

Firmware/odrive-interface.yaml
rowanG077: Fixed.

Firmware/odrive-interface.yaml
rowanG077: Fixed.

rowanG077:
@samuelsadok Fixed the stuff you mentioned and rebased again.

@madcowswe Let me know if you find anything else I will fix it ASAP.

samuelsadok (approved):

docs/thermistors.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Introduction
+Thermistors are elements that change their resistance based on the temperature. They can be used to electrically measure temperature. The ODrive itself has thermistors on board near the FETs to ensure that they don't burn themselves out. In addition to this it's possible to connect your own thermistor to measure the temperature of the connected motors. There are two types of thermistors, Negative Temperature Coefficient (NTC) and Positive Temperature Coefficient (PTC). This indicates whether the resistance goes up or down when the temperature goes up or down. The ODrive only supports the NTC type thermistor.
+
+## FET thermistor
+The temperature of the onboard FET thermistors can be read out by using the `odrivetool` under `<axis>.fet_thermistor.temp`. The odrive will automatically start current limiting the motor when the `<axis>.fet_thermistor.config.temp_limit_lower` threshold is exceeded and once `<axis>.fet_thermistor.config.temp_limit_upper` is exceeded the ODrive will stop controlling the motor and set an error. The lower and upper threshold can be changed values but this is not recommended.
```
madcowswe: small grammar edit: "can be changed values but this is not recommended" -> "can be changed, but this is not recommended"

docs/thermistors.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Connecting motor thermistors
+
+To use your own thermistors with the ODrive a few things have to be clarified first. The use of your own thermistor requires one analog input pin. Under `<axis>.motor_thermistor.config` the configuration of your own thermistor is available with the following fields:
+
+* `gpio_pin`: The GPIO input in used for this thermistor.
+* `poly_coefficient_0` to `poly_coefficient_3`: Coefficient that needs to be set for your specific setup more on that in [Thermistor coefficients](#Thermistor coefficients).
+* `temp_limit_lower` and `temp_limit_upper`: Same as the FET temperature limits.
```
madcowswe: "same principle" or "same idea" (so they don't think they need to put the same values)

docs/thermistors.md
```diff
@@ -0,0 +1,35 @@
...
+* `enabled`: Whether this thermistor is enabled or not.
+
+## Voltage divider circuit
+To measure a temperature with a thermistor a voltage divider circuit is used in addition with an ADC. The screenshot below is taken directly from the ODrive schematic.
+
+![Launch Configurations](screenshots/thermistor-voltage-divider.png "Thermistor voltage divider")
+
+The way this works is that the thermistor is connected in series with a known resistance value. By connecting an ADC directly after the thermistor the resistance value can be determined. For further information see [Voltage divider](https://en.wikipedia.org/wiki/Voltage_divider).
+
+To use a thermistor with the ODrive a voltage divider circuit has to be made that uses `VCCA` as the power source with `GNDA` as the ground. The voltage divider output can be connected to a GPIO pin that supports analog input.
```
madcowswe, line 25: add: "While not strictly necessary, it is a good idea to add a capacitor as shown as well. This will help reduce the effect of electrical noise. A value between 470nF and 4.7uF is recommended, and any voltage rating 4V or higher. Put the capacitor physically close to the ODrive."

docs/thermistors.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+## Thermistor coefficients
+Every thermistor and voltage divider circuit is different and thus it's necessary to let the ODrive know how to relate a voltage it measures at the GPIO pin to a temperature. The `poly_coefficient_0` to `poly_coefficient_3` under `<axis>.motor_thermistor.config` are used for this. The `odrivetool` has a convenience function `set_motor_thermistor_coeffs(axis, Rload, R_25, Beta, Tmin, Tmax)` which can be used to calculate and set these coefficients.
+
+* `axis`: Which axis do set the motor thermistor coefficients for (`odrv0.axis0` or `odrv0.axis1`).
+* `Rload`: The Ohm value of the resistor used in the voltage divider circuit. 
+* `R_25`: The resistance of the thermistor when the temperature is 25 degrees celsius. Can usually be found in the datasheet of your thermistor. Can also be measured manually with a multimeter.
+* `Beta`: A constant specific to your thermistor. Can be found in the datasheet of your thermistor.
+* `Tmin` and `Tmax`: The minimum and maximum temperature limit that is used to create the coefficients. Make sure to set these values that are lower and higher respectively then what is expected during operation.
```
madcowswe: "The temperature range that is used to create the coefficients. Make sure to set this range to be wider than what is expected during operation. A good example may be -10 to 150."

Firmware/MotorControl/motor.cpp
```diff
@@ -186,10 +162,15 @@ float Motor::effective_current_lim() {
     } else {
         current_lim = std::min(current_lim, axis_->motor_.current_control_.max_allowed_current);
     }
-    // Thermal limit
-    current_lim = std::min(current_lim, thermal_current_lim_);
 
-    return current_lim;
+    // Apply axis current limiters
+    for (const CurrentLimiter* const limiter : axis_->current_limiters_) {
+        current_lim = std::min(current_lim, limiter->get_current_limit(config_.current_lim));
```
madcowswe, line 168: I found it a bit confusing that get_current_limit takes an argument. I think the purpose being a "default" if it isn't enabled? If so then this min is redundant right?
Supposing I understand correctly, then I think it might be clearer for get_current_limit to not take any argument and return an optional https://en.cppreference.com/w/cpp/utility/optional

Firmware/MotorControl/thermistor.hpp
```diff
@@ -0,0 +1,74 @@
...
+                             const bool& enabled);
+
+    void update();
+    bool do_checks();
+    float get_current_limit(float user_limit) const override;
+
+    uint16_t adc_channel_;
+    const float* const coefficients_;
+    const size_t num_coeffs_;
+    float temp_;
```
madcowswe: just "temp" by itself can look like "temporary". I would suggest using "temperature" in this case. Also applies to the interface yaml.

madcowswe (commented):
Great work, looks very nice in general.
I have some small comments on some wording, and one suggested code change.

Thanks for your work on this!

Firmware/MotorControl/motor.cpp
rowanG077: The current limiter doesn't "know" the configured current limit. The purpose of the argument is not only being a default but also what base current limit to apply the limit to. An alternative would be for current limiters to not return a concrete current value but rather a current scale between 0.0 and 1.0. And then the logic would simply be to find the smallest scaler and apply it to the configured current limit.

An optional would be overkill I think because we would need the `min` call anyway. Suppose we have two current limiters with a configured limit of 20A.

1. First current limiter says current limit is 15A.
2. Second current limiter says current limit is 12A.

Then we would still need to find the minimum of all the current limits to know which to actually apply. Introducing an optional is simply another case without a (to me) clear benefit. 

Anyway the best way forward I think would be to rewrite this to a scaler based implementaiton. `get_current_limit` would not have any arguments and return a float between 0.0 and 1.0. The `std::min` would remain and the `effective_current_lim_` will become `current_lim * current_lim_scaler`.

madcowswe: Ah ok, the value is also used as a base current limit to scale from, I see.
Now that I understand what is going on, I think it is good as is. Changing it to using `current_lim_scaler` will preclude other limiting scenarios that are an absolute limit rather than ratiometric de-rating.

Perhaps we can rename it to `base_current_lim` instead of user limit, it may make it a bit clearer that the main purpose is to do scaling off, rather than a user default (which is what I misinterpreted it as).

Firmware/MotorControl/thermistor.hpp
rowanG077: Fixed.

Firmware/MotorControl/motor.cpp
rowanG077: Fixed.

docs/thermistors.md
rowanG077: Fixed.

docs/thermistors.md
rowanG077: Fixed.

docs/thermistors.md
rowanG077: Fixed.

docs/thermistors.md
rowanG077: Fixed.

rowanG077:
@madcowswe I fixed your comments and rebased to the latest version.

madcowswe:
Amazing, great work, thank you!

PR #421, opened 2020-06-06
Bulk Capture Python Helper Methods

kylebme:
Added a BulkCapture class for asynchronous capture of data. Also added a step response helper function that does a bulk capture, gives a step position command, and plots the results.

Changed default liveplotter data_rate back to 100.


tools/odrive/utils.py
```diff
@@ -118,11 +118,106 @@ def did_close(evt):
...
+            print("Achieved average data rate: {}Hz".format(total_samples / self.data.iloc[-1, 0]))
+            print("If this rate is significantly lower than what you specified, consider lowering it below the achieved value for more consistent sampling.")
+            self.event.set() # tell the main thread that the bulk capture is complete
+        Thread(target=loop, daemon=True).start()
+    
+    def plot_data(self):
+        import matplotlib.pyplot as plt
+        plt.plot(self.data[0], self.data.drop(0, axis=1))
+        plt.xlabel("Time (seconds)")
+        plt.ylabel("Counts")
```
samuelsadok: This unit would depend on what variable is plotted so I think it's better not to label the Y axis.

tools/odrive/utils.py
```diff
@@ -118,11 +118,106 @@ def did_close(evt):
...
     #plot_data()
 
+
+class BulkCapture:
+    '''
+    Asynchronously captures a bulk set of data when instance is created.
+
+    get_var_callback: a function that returns the data you want to collect (see the example below)
+    data_rate: Rate in hz
+    length: Length of time to capture in seconds
```
samuelsadok, line 143: I'd call this `duration` to distinguish it more clearly from something like "buffer length".

tools/odrive/utils.py
```diff
@@ -118,11 +118,106 @@ def did_close(evt):
...
+        from threading import Event, Thread
+        import pandas as pd
+
+        self.event = Event()
+        def loop():
+            vals = []
+            start_time = time.monotonic()
+            total_samples = int(length * data_rate)
+            period = 1.0 / data_rate
+            for i in range(total_samples):
```
samuelsadok: You could directly use `time.monotonic() - start_time > length` as termination condition to honor the `length` parameter more reliably.

tools/odrive/utils.py
```diff
@@ -118,11 +118,106 @@ def did_close(evt):
...
+                    data = get_var_callback()
+                except Exception as ex:
+                    print(str(ex))
+                    print("Waiting 1 second before next data point")
+                    time.sleep(1)
+                    continue
+                relative_time = time.monotonic() - start_time
+                vals.append([relative_time] + data)
+                time.sleep(period - (relative_time % period)) # this ensures consistently timed samples
+            self.data = pd.DataFrame(vals) # A lock is not really necessary due to the event
```
samuelsadok: In order to keep dependencies low I think we should expose either the plain python array here or use numpy (which we already use in other places) instead of pandas.

The plotting command below would then become `plt.plot(data[:,0], data[:,1:])`.

samuelsadok (commented):
Thanks for the PR! Here are just a couple of minor comments.

kylebme:
Thanks for all your comments. This is great feedback. I've implemented the changes and will request re-review. I also added a title to the plot based on the code used to create the bulk capture.

samuelsadok (approved):

PR #426, opened 2020-06-16
Change motor control input from Amps to Nm

PAJohnson:
The purpose of this PR is to change from units of Amps to units of Nm for the input value to the motor control. This affects multiple parts of the ODrive code structure, including the communication interfaces.


Firmware/MotorControl/controller.cpp
```diff
@@ -153,29 +153,29 @@ bool Controller::update(float* current_setpoint_output) {
...
         case INPUT_MODE_VEL_RAMP: {
             float max_step_size = std::abs(current_meas_period * config_.vel_ramp_rate);
             float full_step = input_vel_ - vel_setpoint_;
             float step = std::clamp(full_step, -max_step_size, max_step_size);
 
             vel_setpoint_ += step;
-            current_setpoint_ = (step / current_meas_period) * config_.inertia;
+            torque_setpoint_ = (step / current_meas_period) * config_.inertia;
         } break;
         case INPUT_MODE_CURRENT_RAMP: {
```
madcowswe: We should also rename this to Torque ramp.

Firmware/MotorControl/motor.cpp
```diff
@@ -177,19 +177,19 @@ bool Motor::do_checks() {
     return true;
 }
 
-float Motor::effective_current_lim() {
+float Motor::effective_torque_lim() {
```
madcowswe: Hm actually I think all the limits involved here make more sense to be in terms of current. I would keep this function as effective_current_lim, and apply the conversion at every applicable call site.

Firmware/MotorControl/motor.hpp
```diff
@@ -29,6 +29,7 @@ class Motor : public ODriveIntf::MotorIntf {
         float Id_measured; // [A]
         float I_measured_report_filter_k;
         float max_allowed_current; // [A]
+        float max_allowed_torque; // [Nm]
```
madcowswe: "max allowed current" is a hardware limit, as noted earlier, probably best to keep the current limits as current and not torque.

Firmware/communication/ascii_protocol.cpp
```diff
@@ -108,15 +108,15 @@ void ASCII_protocol_process_line(const uint8_t* buffer, size_t len, StreamSink&
...
             axis->controller_.input_pos_updated();
             axis->watchdog_feed();
         }
 
     } else if (cmd[0] == 'q') { // position control with limits
         unsigned motor_number;
-        float pos_setpoint, vel_limit, current_lim;
-        int numscan = sscanf(cmd, "q %u %f %f %f", &motor_number, &pos_setpoint, &vel_limit, &current_lim);
+        float pos_setpoint, vel_limit, torque_lim;
+        int numscan = sscanf(cmd, "q %u %f %f %f", &motor_number, &pos_setpoint, &vel_limit, &torque_lim);
```
madcowswe, line 119: Here is a good example where we do in fact want a torque limit (not current limit), as it's a proxy for an acceleration limit. I guess we should have both a torque and current limit be observed seperately.

docs/getting-started.md
```diff
@@ -346,9 +346,9 @@ Set the velocity ramp rate (acceleration): `axis.controller.config.vel_ramp_rate
 Activate the ramped velocity mode: `axis.controller.config.input_mode = INPUT_MODE_VEL_RAMP`.<br>
 You can now control the velocity with `axis.controller.input_vel = 5000` [count/s].
 
-### Current control
+### Torque control
 Set `axis.controller.config.control_mode = CONTROL_MODE_CURRENT_CONTROL`.<br>
-You can now control the current with `axis.controller.input_current = 3` [A].
+You can now control the torque with `axis.controller.input_torque = 3` [Nm].
```
madcowswe: 3Nm is very large (comared to 3 A on common motors). Perhaps use 0.1Nm in the example.

tools/.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
             "type": "python",
             "request": "launch",
             "stopOnEntry": true,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
```
madcowswe: Is this change intended? What is the reason it is changed (I don't know what either does)

Firmware/MotorControl/motor.cpp
```diff
@@ -440,13 +440,14 @@ bool Motor::FOC_current(float Id_des, float Iq_des, float I_phase, float pwm_pha
...
+    float current_setpoint = torque_setpoint / config_.torque_constant;
     current_setpoint *= config_.direction;
     phase *= config_.direction;
     phase_vel *= config_.direction;
 
     // TODO: 2-norm vs independent clamping (current could be sqrt(2) bigger)
-    float ilim = effective_current_lim();
+    float ilim = effective_torque_lim() / config_.torque_constant;
     float id = std::clamp(current_control_.Id_setpoint, -ilim, ilim);
     float iq = std::clamp(current_setpoint, -ilim, ilim);
```
madcowswe: Note that `iq = torque_setpoint / torque_constant` doesn't work for ACIM motors. In that case you need to do:
`iq = torque_setpoint / (torque_constant * acim_rotor_flux)`. Beware of dividing by zero acim_rotor_flux, you can use `max(acim_rotor_flux, acim_gain_min_flux)`

Firmware/MotorControl/controller.cpp
PAJohnson: I think you meant this in reference to INPUT_MODE_CURRENT_RAMP. I changed that to INPUT_MODE_TORQUE_RAMP.

tools/.vscode/launch.json
PAJohnson: This change was not intended and I also do not know what it does.

Firmware/MotorControl/motor.hpp
```diff
@@ -159,6 +162,7 @@ class Motor : public ODriveIntf::MotorIntf {
     } gate_driver_exported_;
     DRV_SPI_8301_Vars_t gate_driver_regs_; //Local view of DRV registers (initialized by DRV8301_setup)
     float thermal_current_lim_ = 10.0f;  //[A]
+    float thermal_torque_lim_ = 10.0f; //[Nm]
```
madcowswe: This variable is not used.

Firmware/MotorControl/motor.cpp
```diff
@@ -440,11 +440,19 @@ bool Motor::FOC_current(float Id_des, float Iq_des, float I_phase, float pwm_pha
 }
 
 
-bool Motor::update(float current_setpoint, float phase, float phase_vel) {
-    current_setpoint *= config_.direction;
+bool Motor::update(float torque_setpoint, float phase, float phase_vel) {
+    float current_setpoint;
```
madcowswe: Maybe set to 0 by default to avoid magic smoke in case of a code error in the future.

Firmware/MotorControl/motor.hpp
```diff
@@ -45,12 +45,15 @@ class Motor : public ODriveIntf::MotorIntf {
         float resistance_calib_max_voltage = 2.0f; // [V] - You may need to increase this if this voltage isn't sufficient to drive calibration_current through the motor.
         float phase_inductance = 0.0f;        // to be set by measure_phase_inductance
         float phase_resistance = 0.0f;        // to be set by measure_phase_resistance
+        float torque_constant = 1.0f;         // to be set by user
```
madcowswe: Units: Nm/A for PM motors, Nm/A^2 for induction motors.

Firmware/MotorControl/controller.cpp
```diff
@@ -316,36 +316,35 @@ bool Controller::update(float* current_setpoint_output) {
...
     } else {
         if (limited) {
             // TODO make decayfactor configurable
-            vel_integrator_current_ *= 0.99f;
+            vel_integrator_torque_ *= 0.99f;
         } else {
-            vel_integrator_current_ += ((vel_integrator_gain * gain_scheduling_multiplier) * current_meas_period) * v_err;
+            vel_integrator_torque_ += ((vel_integrator_gain * gain_scheduling_multiplier) * current_meas_period) * v_err;
         }
     }
```
madcowswe: Since this no longer considers all kinds of current/torque limits (just the straight torque one), we are not implementing the anti-windup correctly. We need to consider all saturation effects on the "actuator" i.e. torque production.
I will follow up on Discord with some theory.

Firmware/MotorControl/motor.cpp
```diff
@@ -192,6 +192,21 @@ float Motor::effective_current_lim() {
     return current_lim;
 }
 
+float Motor::max_available_torque() {
+    //return the maximum available torque for the motor.
+    //Note - for ACIM motors, available torque is allowed to be 0.
+    if (config_.motor_type == Motor::MOTOR_TYPE_ACIM) {
+        float max_torque = effective_current_lim() * config_.torque_constant * current_control_.acim_rotor_flux;
```
madcowswe, line 199: While not a desired operating point, it can happen that acim_rotor_flux is negative. In that case we probably want to clamp it to 0 to avoid issues with reporting negative max_torque.

Firmware/MotorControl/motor.cpp
```diff
@@ -192,6 +192,21 @@ float Motor::effective_current_lim() {
     return current_lim;
 }
 
+float Motor::max_available_torque() {
+    //return the maximum available torque for the motor.
+    //Note - for ACIM motors, available torque is allowed to be 0.
```
madcowswe: By convention we usually put general function documentation right above the function definition rather than right below.

PR #427, opened 2020-06-20
Allow reversal of homing direction and ability to clear SPI Error Rate

camrbuss:
If a user is using SPI based encoders they are currently unable to reverse the direction of homing to the minimum endstop with a proper offset. The following allows the endstop check to allow both sign velocities. 

If a user is using SPI based encoders that provide inconsistent data on startup they may receive `ERROR_ABS_SPI_COM_FAIL` that is cleared on calls to `axis.clear_errors()`, but this error will immediately raised again as the `encoder.spi_error_rate_` has no way to be reset. The following resets `spi_error_rate` upon `axis.clear_errors()`.

### DIFF
--- a/Firmware/MotorControl/axis.cpp
+++ b/Firmware/MotorControl/axis.cpp
@@ -173,10 +173,9 @@ bool Axis::do_checks() {
     // controller_.do_checks();
 
     // Check for endstop presses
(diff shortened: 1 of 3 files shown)


samuelsadok:
Good catch, thanks!

Issue #420, opened 2020-06-06
Stuck in DRV8301_readSpi infinite loop when using absolute encoders


jhurliman:
Here's my setup: ODrive 3.6-56V with AMT203-V absolute encoders with chip select pins connected to GPIO pins 7 and 8 for axis 0 and 1 respectively, running the latest firmware from `devel` (essentially the 0.5.0-rc). I was able to flash the new firmware, run the latest odrivetool, set the encoder CPR to 4096 and encoder mode to `SpiAbsCui` for axis0, then run motor calibration, encoder offset calibration, and go into closed loop control mode. Everything seemed great. However, when I powered the device off then powered it back on again, it refused to respond to requested_state commands. I flashed a debug version of the firmware and connected via ST-LINK, and found it was getting stuck on the following line in `drv8301.c`:

```
while((DRV8301_readSpi(handle,DRV8301_RegName_Status_1) & DRV8301_STATUS1_FAULT_BITS) != 0);
```

The value 0xFFFF was being returned each time, causing that check to fail.

If I set the encoder mode to 0 (Incremental) for both axes, save_configuration(), and reboot() I can get past that. Then, I can set the encoder mode back to 0x100 (SpiAbsCui) and set requested_state to AXIS_STATE_ENCODER_OFFSET_CALIBRATION. Then finally, closed loop control. But if I leave the encoder mode as SpiAbsCui and power cycle, I'm stuck back in that same startup loop.

samuelsadok:
For  `encoder.config.mode` to take effect you _need_ to reboot actually. Then, after the reboot, you need to run offset calibration once and save it. See [here](https://github.com/madcowswe/ODrive/blob/rc-v0.5.0/docs/encoders.md).

Do you have the incremental output of the encoder also connected to the ODrive? If so, I think what's happening is that your encoder is just being used as a normal incremental encoder until reboot. For an SPI encoder you don't need to connect the A/B/Z lines at all.

The other thing is, looking at the datasheet, the protocols for [CUI AMT20x](https://www.cuidevices.com/product/resource/amt20.pdf) and [CUI AMT23x](https://www.cuidevices.com/product/resource/amt23.pdf) seem incompatible. The latter one is implemented in `rc-v0.5.0` while the former is not. Feel free to open a feature request issue for the AM20x (or better yet a PR).

[...]

PR #431, opened 2020-06-30
Lowlevel Refactor

samuelsadok:
Summary of the most important changes:

 - Replace MotorControl/board_config_v3.h by Board/v3/board.cpp (hardware specific variable instantiations and functions) and Board/v3/Inc/board.h (hardware specific declarations and defines)
 - Move the scaffolding of the boot sequence from Board/v3/Src/main.c and Board/v3/Src/freertos.c to MotorControl/main.cpp (including the `main()` entry point).
 - Statically allocate Axis objects
 - Overhaul NVM code to allow for cleaner code organization and facilitate (future) dynamically sized config items.
 - Introduce SPI arbiter to allow multiple components to share the same SPI in a thread-safe way.
 - Introduce explicit GPIO modes in order to avoid unclear and fragile priorities that were a side effect of the init order.
 - Treat A/B/Z encoder I/O and on-board CAN lines as GPIO9...GPIO16. This opens up the possibility to disable the incremental encoders and use their pins for other purposes (e.g. I2C).

HWIL test log:
 171dab31bd3c241c679119813a6b88fd591827bc: all tests pass
 4b0a6fa478d7ba7b7bb318f110c863887c696e43: all tests pass (`analog_input_test.py`, `calibration_test.py`, `can_test.py`, `closed_loop_test.py`, `encoder_test.py`, `fibre_test.py`, `integration_test.py`, `nvm_test.py`, `pwm_input_test.py`, `step_dir_test.py`, `uart_ascii_test.py`)


jhurliman:
Is this ready for testing? I'm wondering if it will solve our issues of the AMT232B and DRV8301 fighting with each other.

jhurliman:
I tried merging this branch into devel and building/flashing it. The device gets stuck in the init sequence just like https://github.com/madcowswe/ODrive/issues/420 except now it happens with or without pull-up resistors.

samuelsadok:
Yes this is ready for testing.
However I wouldn't expect that it solves your issue; I don't think there's a software-caused SPI collision on `devel` because of the order in which things are initialized. I think we need to get to understand your issue separately over at #420.

Firmware/Board/v3/board.cpp
```diff
@@ -0,0 +1,416 @@
...
+        encoders[1], // encoder
+        sensorless_estimators[1], // sensorless_estimator
+        controllers[1], // controller
+        fet_thermistors[1], // fet_thermistor
+        motor_thermistors[1], // motor_thermistor
+        motors[1], // motor
+        trap[1], // trap
+        endstops[2], endstops[3], // min_endstop, max_endstop
+    },
+}};
```
madcowswe, line 139: is it possible to do the struct initialization by name thing like `.stuff = bla` ?

Firmware/Board/v3/board.cpp
```diff
@@ -0,0 +1,416 @@
...
+    {GPIOC, GPIO_PIN_9}, // ENC0_Z
+    {GPIOB, GPIO_PIN_6}, // ENC1_A
+    {GPIOB, GPIO_PIN_7}, // ENC1_B
+    {GPIOC, GPIO_PIN_15}, // ENC1_Z
+    {GPIOB, GPIO_PIN_8}, // CAN_R
+    {GPIOB, GPIO_PIN_9}, // CAN_D
+};
+#else
+#error "unknown GPIOs"
+#endif
```
madcowswe: Please add a comment next to each one so it's easy for human to see which gpio is what at a glance.

madcowswe (commented):
I don't think I will be able to make time to review the rest, just left two basic comments.
Just make sure that the test coverage is decent, then you can just merge it in.

Firmware/Board/v3/board.cpp
samuelsadok: Unfortunately no because these are initializer lists that get applied as arguments to the `Axis` constructor.

PR #436, opened 2020-07-09
Allow motors with attached gearbox find idx

InNoobWeTrust:
With attached gearbox, the default finish distance might not enough and
motor driver can fail to find index position. Expose finish_distance
property to remote object so controller can override.


InNoobWeTrust:
Oops, intended to test this draft on forked repo but mistakenly push PR to upstream, sorry. 😶

PR #437, opened 2020-07-11
Call Encoder::setup() before Motor::setup()

jhurliman:
`abs_spi_cs_pin_init()` sets `GPIO_PULLUP` on the configured CS pins. Without this, using AMT2xx encoders on GPIO pins 7 and 8, `DRV8301_readSpi()` will always read a value of `0xFFFF` causing `DRV8301_enable()` to get stuck in an infinite loop. This PR switches the order of operations to configure CS GPIO pins before attempting to communicate with the DRV8301. In my testing this fixes https://github.com/madcowswe/ODrive/issues/420

### DIFF
--- a/Firmware/MotorControl/main.cpp
+++ b/Firmware/MotorControl/main.cpp
@@ -245,13 +245,13 @@ int odrive_main(void) {
     // must happen after communication is initialized
     pwm_in_init();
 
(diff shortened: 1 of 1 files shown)


jhurliman:
I spoke too soon. The issue is still occurring after this change.

PR #439, opened 2020-07-14
Added pos_abs_ latch in encoder::update()

Zytrahus:
Fix drifting taking place with absolute encoders if the spi callback is fired while encoder::update() is running which would result in shadow_count_ drifting away from count_in_cpr_
Suggested by @PAJohnson on discord on 7/6.


samuelsadok:
Looks good to me, thanks!

PR #440, opened 2020-07-19
Changed bash shebangs to a more general version.

rowanG077:
Some unix operating systems(including mine) do not have bash under `/bin/bash`. This modifies the shebang to lookup bash in the environment.  This is the most portable way I know of.

### DIFF
--- a/Firmware/build.sh
+++ b/Firmware/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Builds the firmware with the configuration specified by
 # environment variables named CONFIG_...
 # If DEPLOY is set, the deliverables are copied to Firmware/deploy/*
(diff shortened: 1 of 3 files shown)


kalvdans (commented):
Just for information, what is your distro and where is your bash?

rowanG077:
I use NixOS. bash is located in the nix-store under an ephemeral hash code. For me it's `/nix/store/16wisfxggxmlz3wasnqp4isp2kd82vbl-bash-interactive-4.4-p23/bin/bash`.

kalvdans:
Thanks! And I presume no one has an experimental `env`. lgtm!

rowanG077:
Using `#!/usr/bin/env bash` is the canonical way afaik for all unix based distros. See [wikipedia](https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability), this includes NixOS. For most other Linux distros the current shebang should work. NixOS is the odd one out.

samuelsadok:
It's my understanding too that `#!/usr/bin/env bash` is the canonical way but I didn't know a distribution exists where `#!/bin/bash` doesn't work. Fwiw those scripts are not expected to be used by most contributors. In fact `build.sh` is deprecated in favor of `.github/workflows/compile.yaml` and `test_runner.py` probably has more issues than that if you run it on something that is not the Raspbian based test rig we're using.

Issue #438, opened 2020-07-13
USB responses over 64 bytes are silently dropped or cause buffer overflow


durka:
This bug has multiple layers which hide each other, but I am fairly confident it's a real bug.

A loop in `interface_usb.cpp` attempts to send everything passed to it, even if the passed-in buffer is larger than a USB packet:
https://github.com/madcowswe/ODrive/blob/306000f46e5d04f21177ee05fcd11469e5282295/Firmware/communication/interface_usb.cpp#L60-L69

However, because the original `length` is passed through on line 63 instead of the newly calculated `chunk`, the loop fails in its stated purpose.

The inner type, `USBSender`, simply returns -1 if asked to send more than 64 bytes, so the loop would exit on the first iteration if `length > 64`. (If `length` is precisely 64, we get through, but I think that's not really intended and the comparison on line 62 ought to be `<=`.)

As it is, the `respond` function in `ascii_protocol.cpp` tries not to send more than 64 bytes anyway (not sure if anyone else, like fibre, uses `TreatPacketSinkAsStreamSink`, I only use the ASCII protocol), so the bug probably won't come up. But, that's not airtight for two reasons:
1. The buffer in `respond` is declared as size 64, but not using the `USB_TX_DATA_SIZE` definition, so could easily get out of sync with that
2. I say "*tries* not to send more than 64 bytes" because `respond` uses `snprintf` incorrectly. From the man page:
  > The functions *snprintf()* and *vsnprintf()* do not write more than `size` bytes (including the terminating null byte ('\0')).  If the output was truncated due to this limit, then the return value is the number  of  characters  (excluding  the  terminating null byte) which would have been written to the final string if enough space had been available.  Thus, a return value of `size` or more means that the output was truncated.

So if one of the `snprintf`s in `respond` *would* write more than 64 bytes, it does not, but `len` is incremented anyway. Then a subsequent `snprintf` or the USB sending code will read/write past the end of the buffer.

madcowswe:
Thanks for the detailed bug report!

samuelsadok:
Thanks for the well-structured bug report!

These are valid concerns and I have addressed them in #441. I believe there is, in released firmware, no actual way to trigger them (the longest response line is 50 characters) but dormant bugs are still bad of course. While looking into this I actually came across another previously dormant bug that awoke just recently (see PR).

Feel free to take a look and comment, otherwise I'll go on merging it in a bit.

[...]

PR #441, opened 2020-07-20
Fix protocol bugs

samuelsadok:
- Fix incorrect use of snprintf (#438)
 - Fix TreatPacketSinkAsStreamSink::process_bytes() quitting when being passed more than 64 bytes (#438)
 - Fix motor subtree being unreachable from ASCII protocol (triggered by the `motor_thermistor` object that was added in #419)

### DIFF
--- a/Firmware/communication/ascii_protocol.cpp
+++ b/Firmware/communication/ascii_protocol.cpp
@@ -36,14 +36,16 @@ static Introspectable root_obj = ODriveTypeInfo<ODrive>::make_introspectable(odr
 // @brief Sends a line on the specified output.
 template<typename ... TArgs>
 void respond(StreamSink& output, bool include_checksum, const char * fmt, TArgs&& ... args) {
-    char response[64];
+    char response[64]; // Hardcoded max buffer size. We silently truncate the output if it's too long for the buffer.
     size_t len = snprintf(response, sizeof(response), fmt, std::forward<TArgs>(args)...);
+    len = std::min(len, sizeof(response));
     output.process_bytes((uint8_t*)response, len, nullptr); // TODO: use process_all instead
     if (include_checksum) {
         uint8_t checksum = 0;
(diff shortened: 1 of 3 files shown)


Firmware/communication/ascii_protocol.cpp
```diff
@@ -36,14 +36,16 @@ static Introspectable root_obj = ODriveTypeInfo<ODrive>::make_introspectable(odr
...
     size_t len = snprintf(response, sizeof(response), fmt, std::forward<TArgs>(args)...);
+    len = std::min(len, sizeof(response));
     output.process_bytes((uint8_t*)response, len, nullptr); // TODO: use process_all instead
     if (include_checksum) {
         uint8_t checksum = 0;
         for (size_t i = 0; i < len; ++i)
             checksum ^= response[i];
         len = snprintf(response, sizeof(response), "*%u", checksum);
+        len = std::min(len, sizeof(response));
         output.process_bytes((uint8_t*)response, len, nullptr);
```
durka, line 49: I'm not entirely clear on the USB middleware so this is kind of theoretical. But does sending stuff in three separate `process_bytes` calls slow things down? I was analyzing some data dumps and found many cases where checksums and line breaks showed up in separate USB reads on the host side.

durka:
I think this addresses my concerns from #438. Added one other semi-related question above.

Firmware/communication/ascii_protocol.cpp
samuelsadok: Yes if the packets are not full (<64B) this slows things down. If throughput becomes a problem for you then combining those three `process_bytes` calls into one could be an effective workaround.

I say workaround because the proper fix would be for the `TreatPacketSinkAsStreamSink` class to coalesce multiple `process_bytes` calls into a single `process_packet` but I wouldn't bother about that right now because the USB code needs an overhaul in general.

Issue #428, opened 2020-06-20
UART communication from Arduino to ODrive delay


Peter-haotian-cui:
Hi, 
I'm using a stm32duino to communicate with ODrive and found a large delay to get a response (almost 1ms). The stm32 UART baudrate is set to 1Mbps, and using a self build ODrive with the same baudrate. Because I want to have a high frequency control (>500Hz), this 1ms delay is too much for my application. Does anyone knows about the reason and how to reduce the response delay.
![!Screenshot_2020-05-30-13-05-52](https://user-images.githubusercontent.com/38802432/85207049-c0b80200-b358-11ea-8a62-09da2d355b62.png)

Thanks!

samuelsadok:
That might be because of this delay: https://github.com/madcowswe/ODrive/blob/master/Firmware/communication/interface_uart.cpp#L89

This delay exists because otherwise the loop would be a busy spin and starve other threads (such as motor control). There's a PR #393 which changes the UART code to use the RX interrupt instead of polling, maybe this fixes your issue.

samuelsadok:
I'm gonna close this now that we have an increased polling frequency but if it's still an issue feel free to reopen.

EMHacker-alt:
This is great !!!

PR #449, opened 2020-08-07
Increase UART polling frequency from 1kHz to 8kHz

samuelsadok:
Addresses #428.

At 1Mbaud/s and 8kHz polling frequency a maximum of 12.5 bytes can arrive per polling interval.
It is therefore unlikely that a higher polling frequency than 8kHz would add significant additional benefit.

### DIFF
--- a/Firmware/MotorControl/axis.hpp
+++ b/Firmware/MotorControl/axis.hpp
@@ -163,6 +163,10 @@ public:
             // TODO: change arming logic to arm after waiting
             bool main_continue = update_handler();
 
+            if (axis_num_ == 0) {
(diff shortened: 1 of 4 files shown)


madcowswe:
Looks good. Please confirm working on the relevant HWIL tests, and if so you can merge it.

samuelsadok:
Ah yeah forgot to add:

HWIL test log:
6fa76e1de802f82b946c9c3cd74deba9070144e8: `uart_ascii_test.py` & `calibration_test.py` pass. `closed_loop_test.py` passes apart from `TestRegenProtection` which for me also fails on `devel` (=> Discord).

PR #451, opened 2020-08-19
Mechanical brake support

aprgl:
Hey! I'm loving this ODrive! I'm hacking on something a bit different and needed a mechanical brake feature, hopefully it could be useful to others. I'm having some issues getting the 'engage' and 'release' functions to show up in the odrivetool, but otherwise it seems to be working. I'm not familiar with yaml so it may be a simple fix to add the function calls?

I thought it would fit better in the low level 'safety' functions but I didn't want to create a bunch of interdependencies.

I tested the functionality on some orange hardware, save_config/recall functions as well. I saw some strange toggling behavior on GPIO1 during reboot which I didn't dig too far into, not sure if that's a known issue.

Cheers,
Shaun & FRED

Oskar edit: video of feature in action - https://youtu.be/6Z1DGk_AfH0


aprgl:
I refactored the code as per the upstream changes - I'm not near hardware to sanity check at the moment though.

Firmware/MotorControl/mechanical_brake.cpp
```diff
@@ -0,0 +1,23 @@
...
+void MechanicalBrake::engage() {
+    get_gpio(config_.gpio_num).write(config_.is_active_low ? 0 : 1);
+}
+
+void MechanicalBrake::release() {
+    get_gpio(config_.gpio_num).write(config_.is_active_low ? 1 : 0);
+}
+
+void MechanicalBrake::set_enabled(bool enable) {
+    if (config_.gpio_num != 0) {
```
samuelsadok: GPIO0 will be a vaild GPIO on ODrive v4. You can remove this check and `gpio.config` will exit safely when called on an inexistent GPIO.

Firmware/MotorControl/mechanical_brake.hpp
```diff
@@ -0,0 +1,28 @@
...
+class MechanicalBrake {
+   public:
+    struct Config_t {
+        uint16_t gpio_num = 0;
+        bool enabled = false;
+        bool is_active_low = true;
+        bool pulldown = true;
+
+        // custom setters
+        MechanicalBrake* parent = nullptr;
```
samuelsadok, line 13: This variable needs to be set manually when loading the config struct.
Here are examples:
https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/axis.cpp#L87
https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/motor.cpp#L65
And `apply_config()` is called here https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/main.cpp#L99-L104

Firmware/odrive-interface.yaml
```diff
@@ -916,6 +916,16 @@ interfaces:
...
+  ODrive.MechanicalBrake:
+    c_is_class: True
+    attributes:
+      config:
+        c_is_class: False
+        attributes:
+          gpio_num: {type: uint16, c_setter: set_gpio_num}
+          enabled: {type: bool, c_setter: set_enabled}
+          is_active_low: bool
+          pulldown: bool
```
samuelsadok: If you want to add the `engage` and `release` functions this should work:
```
ODrive.MechanicalBrake:
  [...]
  functions:
    engage:
    release:
```
In general you'd have `in` and `out` lists for functions but in this case they are empty so they can be omitted (this is not a YAML rule, it's just how the interface generator is implemented).

samuelsadok (commented):
Thanks for your contribution!

Looks clean, I just added a couple of minor comments.

The GPIO1 toggling issue is probably because of [this](https://github.com/madcowswe/ODrive/blob/devel/Firmware/Board/v3/Src/usart.c#L155) which gets called indirectly from [here](https://github.com/madcowswe/ODrive/blob/devel/Firmware/Board/v3/board.cpp#L277). Since we actually init GPIOs directly now according to `config.gpioX_mode` we can remove GPIO init/deinit from CubeMX's autogenerated files (`usart.c`).

Firmware/MotorControl/mechanical_brake.cpp
aprgl: Makes sense, interesting find on the start-up toggle as well.

Firmware/MotorControl/mechanical_brake.hpp
aprgl: Done and done

Firmware/odrive-interface.yaml
aprgl: Yeah that was my expectation too but I kept running into this when I was trying that.
```
$ make                                                                                                                                                                    
 25) [1.093s] python3 -B ../tools/odrive/version.py --output autogen/version.c                                                                                                                             
Firmware version 0.4.12-dev (fw-v0.4.12-721-g87ea4a42*)
 *** tup messages ***
tup warning: Writing to hidden file '/Users/shaun/gits/ODrive/.git/index'
* 21) arm-none-eabi-g++ -std=c++17 -Wno-register -c communication/communication.cpp -DUSB_PROTOCOL_NATIVE -DUART_PROTOCOL_ASCII -DSTM32F405xx -DARM_MATH_CM4 -mcpu=cortex-m4 -mfpu=fpv4-sp-d16 -DHW_VERSION_MAJOR=3 -DHW_VERSION_MINOR=6 -DHW_VERSION_VOLTAGE=56 -D__weak="__attribute__((weak))" -D__packed="__attribute__((__packed__))" -DUSE_HAL_DRIVER -mthumb -mfloat-abi=hard -Wall -Wdouble-promotion -Wfloat-conversion -fdata-sections -ffunction-sections -O2 -ffast-math -fno-finite-math-only -IDrivers/DRV8301 -IMotorControl -Ifibre/cpp/include -I. -Idoctest -IBoard/v3/Middlewares/Third_Party/FreeRTOS/Source/portable/GCC/ARM_CM4F -IBoard/v3/Middlewares/Third_Party/FreeRTOS/Source/include -IBoard/v3/Middlewares/Third_Party/FreeRTOS/Source/CMSIS_RTOS -IBoard/v3/Middlewares/ST/STM32_USB_Device_Library/Core/Inc -IBoard/v3/Middlewares/ST/STM32_USB_Device_Library/Class/CDC/Inc -IBoard/v3/Drivers/STM32F4xx_HAL_Driver/Inc -IBoard/v3/Drivers/STM32F4xx_HAL_Driver/Inc/Legacy -IBoard/v3/Drivers/CMSIS/Device/ST/STM32F4xx/Include -IBoard/v3/Drivers/CMSIS/Include -IBoard/v3/Inc -I.  -o build/obj/communication_communication.cpp.o
In file included from communication/communication.cpp:78:
communication/../autogen/endpoints.hpp: In function 'bool fibre::endpoint_handler(int, fibre::cbufptr_t*, fibre::bufptr_t*)':
communication/../autogen/endpoints.hpp:1428:180: error: invalid static_cast from type 'MechanicalBrake*' to type 'ODriveIntf::MechanicalBrakeIntf*'
 1428 |         case 319: { return odrive_mechanical_brake_engage(static_cast<ODriveIntf::MechanicalBrakeIntf*>(ODriveIntf::AxisIntf::get_mechanical_brake(ODriveIntf::get_axis0(&ep_root))), input_buffer, output_buffer); } break;
      |                                                                                                                                                                                    ^
communication/../autogen/endpoints.hpp:1429:181: error: invalid static_cast from type 'MechanicalBrake*' to type 'ODriveIntf::MechanicalBrakeIntf*'
 1429 |         case 320: { return odrive_mechanical_brake_release(static_cast<ODriveIntf::MechanicalBrakeIntf*>(ODriveIntf::AxisIntf::get_mechanical_brake(ODriveIntf::get_axis0(&ep_root))), input_buffer, output_buffer); } break;
      |                                                                                                                                                                                     ^
communication/../autogen/endpoints.hpp:1662:180: error: invalid static_cast from type 'MechanicalBrake*' to type 'ODriveIntf::MechanicalBrakeIntf*'
 1662 |         case 553: { return odrive_mechanical_brake_engage(static_cast<ODriveIntf::MechanicalBrakeIntf*>(ODriveIntf::AxisIntf::get_mechanical_brake(ODriveIntf::get_axis1(&ep_root))), input_buffer, output_buffer); } break;
      |                                                                                                                                                                                    ^
communication/../autogen/endpoints.hpp:1663:181: error: invalid static_cast from type 'MechanicalBrake*' to type 'ODriveIntf::MechanicalBrakeIntf*'
 1663 |         case 554: { return odrive_mechanical_brake_release(static_cast<ODriveIntf::MechanicalBrakeIntf*>(ODriveIntf::AxisIntf::get_mechanical_brake(ODriveIntf::get_axis1(&ep_root))), input_buffer, output_buffer); } break;
      |                                                                                                                                                                                     ^
 *** tup messages ***
 *** Command ID=605 failed with return value 1
 [ETA=7s  Remaining=18 Active=0 ]  52%                                                                                                                                                                     
 *** tup: 1 job failed.
tup warning: Update resulted in 1 warning
make: *** [all] Error 1
```
I dug through the autogenerated code for a while and I didn't see anything too different being generated by the YAML structures with functions that successfully compiled vs the new one for this function. Does that error mean anything special to you @samuelsadok ?

samuelsadok: Ahh yeah sorry: Now that `MechanicalBrake` is exporting functions it must inherit from the autogenerated base class `ODriveIntf::MechanicalBrakeIntf`. I.e. `class MechanicalBrake : public ODriveIntf::MechanicalBrakeIntf { ...`. For this you'll also need to include `<autogen/interfaces.hpp>` in `mechanical_brake.hpp`. For an example you can look at `motor.hpp`.

aprgl: Ahhh, I knew I must have been missing an include somewhere, just was looking in the wrong direction. Thanks!

aprgl: Ahh, I  see what I was missing now. Thanks for looking into that. The fixes are compiling in the green now - I'll check it one more time on hardware and then push the changes.

aprgl:
Tested good on hardware, however in loading the configuration on boot, the GPIO config looks like it's being overwritten by  https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/main.cpp#L397-L503. Calling `config_apply_all()` after this block, or doing anything which would trigger `mechanical_brake::apply_config()` again would reset the configuration of the GPIO.

aprgl:
I tested this out on hardware tonight and everything was looking good, but on reboot the GPIO configurations in `MechanicalBrake::apply_config()` are overwritten by the [set gpio modes function in main](https://github.com/madcowswe/ODrive/blob/devel/Firmware/MotorControl/main.cpp#L397-L503). The settings save and load correctly, it's just reverting to default pin configs. I looked into adding a `GPIO_MODE_MECH_BRAKE` or even `GPIO_MODE_DIGITAL_OUT` but I think that might not be solving the right problem here. Is it better to have a call to something along the lines of `apply_extended_configs()` after the default pin-mode set-up?

samuelsadok:
Oh yeah this is closely related to #452.
Based on the thoughts outlined in https://github.com/madcowswe/ODrive/issues/452#issuecomment-679925071 I think adding a `GPIO_MODE_MECH_BRAKE` and removing the `config()` call from `MechanicalBrake::set_enabled()` is the cleanest solution.
Btw I see that there's a pullup variable which doesn't really make a difference for a push-pull output so it can be removed.

aprgl:
Tested good on hardware, settings persisted through `save_configuration()` & `reboot()`. I didn't end up removing the call in 'enable' unless the strategy was only to have those configuration changes take effect on reboot? Happy to swap if that's the intent.

aprgl:
Tested good on hardware, hardware configuration persisted though `save_configuration()` and `reboot()`, implemented the above suggestions and removed the internal resistor calls on the outputs.

Firmware/MotorControl/mechanical_brake.cpp
```diff
@@ -0,0 +1,22 @@
...
+}
+
+void MechanicalBrake::release() {
+    get_gpio(config_.gpio_num).write(config_.is_active_low ? 1 : 0);
+}
+
+void MechanicalBrake::set_enabled(bool enable) { 
+    if (enable) {
+        // We need this flag to alert the system to the configuration on boot
+	    odrv.config_.gpio_modes[config_.gpio_num] = ODriveIntf::GPIO_MODE_MECH_BRAKE;
```
samuelsadok: I think it's better not to automatically change the pin config here. If every enabled component changes pins to itself then we're kinda back where we were before where it's undefined, given a certain configuration, which component "wins" the GPIO.

Removing this function also allows us to remove some more code: the `Config_t::set_enabled()` and `Config_t::set_gpio_num()` hooks, consequently the `Config_t::parent` field and, further again, the `MechanicalBrake::apply_config()` function.

Firmware/MotorControl/mechanical_brake.cpp
```diff
@@ -0,0 +1,22 @@
+#include <odrive_main.h>
+
+bool MechanicalBrake::apply_config() {
+    config_.parent = this;
+    set_enabled(config_.enabled);
+    return true;
+}
+
+void MechanicalBrake::engage() {
+    get_gpio(config_.gpio_num).write(config_.is_active_low ? 0 : 1);
```
samuelsadok: Please wrap this inside `if (enabled)` (likewise in `MechanicalBrake::release()`). Otherwise, if the selected gpio is used as an output pin for something else (specifically SPI nCS), there will be unexpected toggling of the pin.

Alternatively you could also wrap it inside `if (odrv.config_.gpio_modes[config_.gpio_num] == ODriveIntf::GPIO_MODE_MECH_BRAKE)` and remove the `enable` field entirely.

docs/mechanical-brakes.md
```diff
@@ -0,0 +1,64 @@
...
+### is_active_low
+Most safety braking systems are active low, e.g. when the power is off, the brake is on. If the system uses brake drive electronics which use active high logic, flip this bit then reconsider the safety implications of your design...
+
+### Example
+
+Let's say we're hacking away on an old ABB robotic arm. We've wired a 24V brake drive circuit triggered by GPIO5. When GPIO5 is driven, it will release the brakes on the axis we're moving.
+
+```
+<odrv>.<axis>.mechanical_brake.config.gpio_num = 5
+<odrv>.<axis>.mechanical_brake.config.enabled = True
```
samuelsadok: Please add: `<odrv>.config.gpio5_mode = GPIO_MODE_MECH_BRAKE`.

samuelsadok (commented):
Thanks, a few more comments but we're almost there.

docs/mechanical-brakes.md
aprgl: Ahh, this is what I was trying to avoid. It's a bit of a bummer to make the user have to manually configure so many settings for each feature but I see what you're going for.

samuelsadok: Yes I agree so the plan is to have this sort of configuration done by the GUI eventually. Think of a picture of the ODrive where you can click on the GPIOs and select "Mechanical Brake" or drag a wire to a Mechanical Brake block. Or some kind of step-by-step setup wizard.

For this reason we're aiming for a low-level machine-to-machine API rather than a machine-to-user API that does things behind the back of the host.

Firmware/MotorControl/mechanical_brake.cpp
aprgl: Clever. I was a bit worried about a strange edge case there potentially triggering on the wrong axis brake, eg configuring both breaks to the same GPIO channel. I think that's a bit in the weeds though.

aprgl:
Did a quick sanity check on hardware, worked as expected.

aprgl:
[Tested it in two channel mode!](https://www.youtube.com/watch?v=NLHGa5abREY)

samuelsadok (approved):
Sweet, looks good to me now, thanks!

madcowswe:
Thank you for your contribution!

PR #482, opened 2020-09-05
Clean math funcs

madcowswe:
Probably merge `faster-fmodf` first.

- change fmaxf to std max
- change fabsf to std abs
- change to cmath by default
- no need for special fma function, the compiler knows
- drop flag fno-finite-math-only, and implement explicit nan check function


Firmware/MotorControl/utils.hpp
```diff
@@ -54,66 +54,231 @@
...
-    return out;
+// Wrap value to range.
+// With default rounding mode (round to nearest),
+// the result will be in range -y/2 to y/2
+inline float wrap_pm(float x, float y) {
+#ifdef FPU_FPV4
+    float intval = (float)round_int(x / y);
+#else
+    float intval = nearbyint(x / y);
+#endif
```
kalvdans, line 108: What is special about FPv4 when it comes to rounding?

PR #491, opened 2020-09-23
CAN Handling Improvements

Wetmelon:
* Use ref or const ref where applicable instead of pointers
* Switch all `get_x_callback` functions to use `can_setSignal` function instead of copying bytes manually
* Add a cyclic message handler
* Add some data types in axis to contain CANConfig and CAN runtime data per axis.
* Remove `msg` from `get_x_callback` functions by moving `if(msg.rtr)` to the switch statement


samuelsadok (approved):

samuelsadok:
81ae829fce71b314d6e85fec698a1641de8d522f passes `can_test.py`

PR #496, opened 2020-10-01
fix commands.md broken api/ links

valentin-martinez-gama:
#Control Mode (line41) and #ControlCommands (line 54) links are not working on Documentation page. Checking the odrive_interface.yaml file there seems to be an unnecessary .axis in the yaml reference for both links.


samuelsadok (approved):
Thanks for the fix!

PR #497, opened 2020-10-01
Update CHANGELOG to mentiond gain unit changes

valentin-martinez-gama:
Changelog didn't mention changes in gains units and steps on how to update previous gains to work with new odrivetool version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Please add a note of your changes below this heading if you make a Pull Request.
 
 ### Changed
 * **`input_pos`, `input_vel`, `pos_estimate_linear`, `pos_estimate_circular`, are now in units of [turns] or [turns/s] instead of [counts] or [counts/s]**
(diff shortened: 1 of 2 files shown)


valentin-martinez-gama:
This is only mentioned in the [Firmware v0.5.1 Release Candidate Feedback Thread](https://discourse.odriverobotics.com/t/firmware-v0-5-1-release-candidate-feedback-thread/5210) under the **Migration Guide** section. 

It is a bit more clear and explanatory than what I came up with in my commits. Perhaps it would be better to commit something along with those words then to the CHANGELOG? 

A quick note however:  Using the suggested migration formula  `(torque_constant = 8.27 / motor KV) * encoder cpr` with the 270Kv Odrive Motor and 8192 encoder gives a migration constant `= ~250` and a 150Kv motor would give a migration constant of `= ~450`.

Is the 320 you used to change the default gains some kind of an in-between?

PAJohnson:
Thanks for catching this! For the "migration constant", use the formula instead. The stock gains use a torque_constant of 0.04 to split the difference between the 270KV motor and 150KV motor so that they are stable out of the box, thus the migration constant is different. There is no single value to multiply the gains by that will work, it is dependent on motor and encoder properties.

Can you change the wording in your change to `CHANGELOG.md` from

"`pos_gain` is not really affected but previous `vel_gain` and `vel_integrator_gain` in A/(counts/s) should be multiplied by 320 to be converted to the new units."

to

"`pos_gain` is not affected. Old values of `vel_gain` and `vel_integrator_gain` should be multiplied by `torque_constant * encoder cpr` to convert from the old units to the new units." ?

valentin-martinez-gama:
Done @PAJohnson. If there is anything else left to do let me know and I'll help or feel free to finish it yourself.  :)

PAJohnson (approved):

PR #499, opened 2020-10-02
Implement GitHub workflow to automate releases

samuelsadok:
The workflow creates a GitHub Release Draft as follows:

 - Title with a version number parsed from the git ref that triggerd the workflow
 - Body parsed from the changelog
 - Compiled .elf and .hex files for board versions 3.5 and 3.6
 - Python package (as .tar.gz file)

The workflow was tested by adding the trigger `on.push.branches: [autorelease/v0.5.2]` and then pushing onto that branch.

Here's a [test release draft](https://github.com/madcowswe/ODrive/releases/tag/untagged-c987189174808a1fb95f) that was created by this workflow (only visible to collaborators).

### DIFF
--- a/.github/actions/release-firmware/action.yml
+++ b/.github/actions/release-firmware/action.yml
@@ -1,0 +1,57 @@
+name: 'Release Firmware'
+description: 'Compiles the firmware for the specified board version and uploads it to a GitHub release url'
+inputs:
+  board_version:
+    required: true
+  url:
+    required: true
+  token:
+    required: true
+runs:
+  using: "composite"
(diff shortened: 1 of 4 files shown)


PAJohnson:
@samuelsadok Any updates on this? I think we still want this to be automated

samuelsadok:
This is done, we can merge it. I kept it open so @madcowswe can have a look at the release draft but it seems the draft has expired meanwhile. But since this GitHub workflow only creates release drafts (not actual releases) we can merge it now and tweak it when we do the next release.

Issue #505, opened 2020-10-12
Thermistor code organization


PAJohnson:
Thermistors are out of place as a top level object. Suggestion: put under Motor.

PAJohnson:
Fixed in PR #507

PR #507, opened 2020-10-14
Change thermistors from top-level to Motor

PAJohnson:
This PR addresses issue #505. In addition, motor thermistors have been broken on devel for some time; this PR fixes that.

Changes:
`fet_thermistor` and `motor_thermistor` are now members of the `Motor` class
`motor_thermistor` requires some initialization - `apply_config()` method added for this purpose
Thermistor errors rolled into motor errors
odrivetool and GUI updated to handle change
Removed CurrentLimiter, Thermistor arrays abstraction. `motor_thermistor` and `fet_thermistor` `get_current_limit()` and `do_checks()` methods are called explicitly.


Firmware/odrive-interface.yaml
```diff
@@ -444,8 +444,6 @@ interfaces:
             # status_reg_2: readonly uint32
             # ctrl_reg_1: readonly uint32
             # ctrl_reg_2: readonly uint32
-      fet_thermistor: OnboardThermistorCurrentLimiter
-      motor_thermistor: OffboardThermistorCurrentLimiter
```
samuelsadok, line 448: Please add a note to "API migration notes" that these objects were moved. The idea is that this section is a place where users can easily spot what's up when they get an "AttributeError" after an firmware upgrade.

Firmware/MotorControl/motor.cpp
```diff
@@ -110,7 +117,16 @@ bool Motor::do_checks() {
         set_error(ERROR_DRV_FAULT);
         return false;
     }
-
+    if (!motor_thermistor_.do_checks()) {
+        axis_->error_ |= Axis::ERROR_OVER_TEMP;
```
samuelsadok: We can use `Axis::ERROR_MOTOR_FAILED` here, the semantic of which is "the Axis stopped because the motor failed, the motor error reveals more details.".

You can leave the OverTemp error code in the YAML (with a comment that it's unused) so that we don't accidentally reuse the same error code in the future.

Firmware/MotorControl/motor.cpp
```diff
@@ -66,6 +70,9 @@ bool Motor::apply_config() {
     config_.parent = this;
     is_calibrated_ = config_.pre_calibrated;
     update_current_controller_gains();
+    fet_thermistor_.motor_ = this;
+    motor_thermistor_.motor_ = this;
```
samuelsadok: This would be more suitable to put in the Motor constructor (it doesn't need to re-run when reloading the config).

Firmware/MotorControl/motor.cpp
```diff
@@ -66,6 +70,9 @@ bool Motor::apply_config() {
...
+    motor_thermistor_.apply_config();
```
samuelsadok: While this is a valid place for this call I would probably rather move it to the top level (`config_apply_all` in main.cpp). The reason is mainly consistency: the related config operations (load, store, clear) are all done directly from the top level too (also for the Motor object, which, to the user, is a subobject of Axis).

Firmware/odrive-interface.yaml
PAJohnson: Good call! I updated the API Migration Notes.

Firmware/MotorControl/motor.cpp
PAJohnson: Addressed in fd848f2

Firmware/MotorControl/motor.cpp
PAJohnson: Addressed in fd848f2

Firmware/MotorControl/motor.cpp
PAJohnson: Addressed in fd848f2

samuelsadok (approved):
Great, looking good now!

PR #510, opened 2020-10-21
Compute the time in milliseconds to decide when to send heartbeat in can

Ibarria:
CAN was checking ticks instead of milliseconds, yet the configuration parameters are all in milliseconds. 

One possible further change (for discussion) would be to remove the conditional of only updating the send time in case of success. If there is a can issue and we could not send, the current code would try in a very tight loop and could cause timing issues elsewhere.

### DIFF
--- a/Firmware/communication/can_simple.cpp
+++ b/Firmware/communication/can_simple.cpp
@@ -304,19 +304,19 @@ int32_t CANSimple::send_heartbeat(const Axis& axis) {
 }
 
 void CANSimple::send_cyclic(Axis& axis) {
-    const uint32_t now = osKernelSysTick();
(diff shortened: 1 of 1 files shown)


Wetmelon:
Thanks for the PR!  The sys tick happens to be 1 ms in our code so it worked but you indeed have a better solution.

As for the CAN re-try , it should only be retrying once per cycle (125 microseconds), so minimal overhead.

PAJohnson:
@Ibarria In the current `devel` branch, CAN uses `HAL_GetTick()` instead of `osKernelSysTick()`. The HAL function should always return ticks in milliseconds. Unless there are other concerns, I'm closing this PR.

PR #513, opened 2020-11-02
Added input_mode enum values

mdhom:
**Current situation**
When adding a control for `odrv0.axisX.controller.config.input_mode`, it is not treated as an enum, so no dropdown control is generated but only a numeric input.

**What this PR changed**
Now the `input_mode` is also treated as an enum and therefore a dropdown control is generated.

image

### DIFF
--- a/GUI/src/views/Dashboard.vue
+++ b/GUI/src/views/Dashboard.vue
@@ -254,6 +254,40 @@ let odriveEnums = {
       text: "Position Control",
       value: 3,
     },
+  ],
+  input_mode: [
+    {
+      text: "Inactive",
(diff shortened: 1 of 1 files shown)


PAJohnson (approved):
Looks good to me! Thanks for submitting this PR.

mdhom:
@PAJohnson thank you for accepting my PR. I have some other ideas for improving the GUI, should i directly program those ideas and create a PR or is there any channel on which to talk about my ideas before starting? Also, i think this PR has fullfilled one task of the GUI project, but i can't move / edit them. Could you please check that?

PR #514, opened 2020-11-02
Introduced new control-type "watch"

mdhom:
**Current situation**
When using odrive, i want to have a live-view on some properties. For that, i added a control for each property but noticed, that the controls value is not updated / refreshed. So the "control"-type is only for changing values in odrive, not for displaying states. When using plots for getting current states, enums can not be handled and it is difficult to visualize lots of different properties in single plots...

**Suggested and implemented solution**
I introduced a new category of controls, called "watch". A watch control is readonly but refreshes its value every second by pulling the value from the odrive. It distinguishes between Boolean, Numeric and Enum types and displays them differently.

image

**Discussion**
- Is there any option for getting value refreshes other than pulling cyclically? How do the plots get their real-time data?
- Has polling a negative influence in e.g. plots performance?
- Should the polling interval be adjustable? E.g. axis state may not be changing as rapidly as mabye set_pos will...


mdhom:
I just realized, that now also the Control controls are refreshed / updated. That is probably not very usable, so i would have to remove that behaviour from the Control controls.

mdhom:
Ok, what do you think about that:
image
Instead of introducing watch types, i enhance to control types as follows:
- the actual value of the property is refreshed once a second
- if the property has write access, an extra control (input text, select, checkbox) is inserted left of the current value
- if this control is changed OR the send button within the two controls is pressed, the value is written down. 
- you get feedback if writing was successfull whenever the current value has changed

I find that approach quite convenient, because the behaviour is obvious and you can prepare multiple controls with different values for fast switching between two values:
image

- [X] Implement that behaviour for CtrlNumeric
- [ ] Implement that behaviour for CtrlEnum
- [ ] Implement that behaviour for CtrlBoolean
- [ ] Implement that behaviour for CtrlSlider

PAJohnson:
I like this idea! I had considered something similar. For the plots: right now, those are managed separately as "sampled variables". Variables added to the plots are added to `store.state.sampledProperties`. When sampling is enabled, an event is sent to the server to begin a polling loop at a 20ms period. You can see that in `odrive_server.py` beginning at line 103.

There may be some new functionality in Fibre to have "push" actions from the ODrive itself on variable changes. For now, polling will have to be used. This should probably be done in a similar way to the plot variables, by setting up another list in the store to maintain low-rate polled variables.

Part of the reason that this is complicated/messy is the python server acting as the link between ODrive USB communications and the GUI. There is work being done right now by @samuelsadok to make Fibre USB work natively in Electron, so the python server will eventually not be used. I'm not sure of the timeline for that, but it is worth mentioning. I think that the same GUI framework can be used (a list of low-rate polled parameters), but we'll just redirect the sampling calls as necessary. Thanks again for working on this!

mdhom:
ok, that sounds like there will be a major redesign, so for until now, i could do the polling from within vue?
I am not sure if i am capable of changing the server.py, as i am not familiar with python...

PAJohnson:
Sure, polling from within Vue is a good approach. In that case, a good pattern to use would be something similar to `Axis.vue`, starting at line 232.

In the created hook, a polling function is declared to call `fetchParam()` and then `getVal()`. fetchParam updates the value of the parameter in `store.state.odrives`, getVal actually gets that value into a variable.

mdhom:
Ok, i'd love to add this feature then. Two things:
```
  created() {
    ...
    let update = () => {
      fetchParam(this.name);
    }
    this.intervalId = setInterval(update, 1000);
    update();
  },
  beforeDestroy() {
      clearInterval(this.intervalId);
  }
```
Regarding polling, i'd prefer a solution like this, with clearing the interval on destroying the component. With the approach in Axis.vue, the timer is started again and again when hot-reloading during development, so after a few times, you will have many intervals running at the same time. Do you consent?

And my other question is, as far as i can see, when fetching a param, a command is sent to the server script to refresh that parameter. The only thing i haven't found yet is, how and where is the actual new value applied to the object in store and how is that then forwarded to the control that sent the request? That looks like magic at the moment to me 😅

The CtrlSlider would look like this:
image
I added the write button on top of the slider, so you are able resend the same value without having to move the slider again. Also, i am now adding the min and max values to the control object which can be exported. So you can import and export these settings and do not have to re-apply them each time you load the dashboard.
```
    {
      "controlType": "CtrlSlider",
      "path": "odrives.odrive0.axis0.controller.config.vel_gain",
      "max": 0.7,
      "min": 0,
      "sliderValue": 0.167
    }
```

mdhom:
The CtrlBoolean would look like this, accordingly. Also, the state of the checkbox is im-/exported with the dashboard:
image

mdhom:
The CtrlEnum would look like this, accordingly. Also, the state of the selectis im-/exported with the dashboard:
image

PAJohnson:
That looks good to me, I will have to do a more thorough review before merging. However, it looks like you are mostly done (I see some commits for each type of control)?

For `fetchParam()` calls - a socketio event is fired by that function in `odrive_utils.js`. The server receives that event, gets the new value of that parameter, and then emits a socketio event called `ODriveProperty` with the relevant data in a dict. On that event, a mutation is triggered in the store to update that specific parameter in `store.state.odrives`. The Vue reactivity system should then update all of the components that make use of that parameter. As long as the Vue reactivity works correctly, it is magic 😄 .

PAJohnson:
Ok, a more thorough review. I see what you were going for with the save button idea, but to me there are two issues. One, the functionality of "flip between multiple values" is handled already by the Action components. An Action for boolean values is missing, but otherwise they already have this functionality (including saving the set value to an exported dashboard). The Controls are somewhat meant to be a transparent view/handle for the ODrive parameter.

Two: I find the duplication of set value, and getVal() value for each control to be visually messy. I'm also not sure if the save button is necessary - If the controls have live updates via the setInterval cal, and the ODrive is not disconnected, you can be quite sure that the value that you entered will make its way to the ODrive. I think that having a live update would be sufficient for this functionality (if you send a value and it "does not take", the setInterval update will set it back to what it actually is). 

So to simplify things, I propose these changes: remove the usage of `parentControl` for the visual elements, because if the Actions are used for the "toggle value" functionality, the regular Controls do not need memory of what they were set to. Remove the duplication of a watch value and <input> element for the various controls, and also the save button. Keep the persistent updates on the controls, and remove "watch" elements (I think that "watch" behavior by default is OK). To maintain feature parity with the PR, could you make an ActionBoolean component? 

This way, there would be a way to have saved values/toggles for all ODrive parameters via the Actions, and the Controls with polled updated would function as Watch components by default. These comments are mostly to do with updated and clarifying components that already exist, and having maximum functionality per pixel on the page.

Also, thanks for fixing the Slider component and cleaning up the code in various places (especially the obvious usage of computed `name` in the other computed properties :smile:)! What do you think about the proposed changes?

PAJohnson:
@mdhom Any updates on the proposed changes? If not, I'd like to close this PR.

PR #515, opened 2020-11-04
Add index_offset as an option

Wetmelon:
Set the index_offset in units of [turns].  When the index is found, set_linear_count is used to adjust the assumed position of the index.


samuelsadok (approved):

Wetmelon:
Rebased to `devel`

PR #516, opened 2020-11-05
Add support for AMT21 encoders

geoffreywwang:
This PR adds support for AMT21 encoders which communicate using UART over RS485.

The current implementation is basic and only works on M1 and only requests data from the default address (0x54). It currently uses USART1 with the M1_ENC_A and M1_ENC_B pins to interface with the encoder. The DE/nRE pin for the RS485 transceiver is controlled by the M1_ENC_Z pin and an interrupt signal once the request transmission is complete.

In the current state, the ODrive can read data from the encoder and seems to update shadow_count and all other encoder counts correctly. Waiting on changes to low level control to be merged in before trying to organize/improve the current implementation.

Potential improvements:
- Add configurable encoder address
- Support using AMT21 encoders on both axes
- Disable using AMT21 encoder on M0 if not enabled on M1
- Leverage multiturn AMT21 encoder data (AMT21x-D)


samuelsadok:
Oh sorry seems this PR got auto-closed because I merged and deleted the `uart1` branch. Would you mind reopening towards `devel`?

geoffreywwang:
Done! #517 created.

pbrier:
Did this feature go somewhere on one of the active branches? Any plans for integration? Would love to test this!

samuelsadok:
We decided to focus on ODrive v4 for AMT21 support because it has an onboard RS-485 transceiver. It is directly connected to the MCU's DE pin which means we can actuate the transceiver's direction directly from the UART hardware. But it also means that the implementation unfortunately doesn't work with v3 hardware.

PR #521, opened 2020-11-10
Convert run_lockin_spin to template function instead

Wetmelon:


samuelsadok:
To avoid moving too much code into headers and templates I would suggest to rather use a "callback + context pointer"-style approach. Something like this: https://godbolt.org/z/a56jb6

PR #528, opened 2020-11-18
Support Infineon TLE5012b abs SPI encoder

jonmendenhall:
This encoder is different than other SPI encoders because it uses an 3-wire SPI-compatible SSC interface.

- CS: same as normal chip select
- SCK: same as normal SPI clock
- DATA: bidirectional data line as opposed to separate MOSI / MISO
(Signal across wires is still the same as SPI)

Because of this, Infineon suggests adding a ~3.3K resistor between MOSI / MISO if you want to work with standard 4-wire SPI, but I didn't want to interfere with any other SPI devices, maybe the DRVs wouldn't be happy about it... or it could have other effects on the SPI bus... not sure.

Anyways, the SPI peripheral on stm32 can be set to use bidirectional mode (SPI_DIRECTION_1LINE) where the MOSI pin can be set to send out data, or receive data so I went with that option. 

The **spi_arbiter** was really the main part that needed to be updated to support these bidirectional SPI tasks, so I added an option to specify if a task is meant to **split_tx_rx** to encompass this.

### Signals
These show the SPI signals coming across the wires for 2 separate encoders, and the code is working fine to send the read angle value command (0x8020), then receiving the value of the angle value register.

I bumped the BaudRatePrescaler from 32 to 8 to get a much faster transaction time because I plan on using 4 of these encoders on the same ODrive (a motor encoder, and joint encoder per axis)... and I actually have that working with closed loop control and a 24:1 gear reduction, but I think I'll work on it a bit more before I open a pull request for it :)

_(My cheap logic analyzer can only sample up to 24Mhz, and the clock pulses are just past the limit, so they appear to have varying widths at around 1-2 samples)_

image

image

### Things to address / investigate
- In the two pictures, you can see that the entire SPI transactions are different lengths, the first is 16us, the second is 12us... I'm guessing this is because other code got scheduled in between the send/receive of the first encoder, while the second encoder transaction didn't get interrupted.
- You'll also see that the SCK line properly goes high after the first receive is complete, but not after the second transaction... I'm guessing this is because I had to disable the SPI peripheral to prevent Overruns (SCK kept pulsing past the specified 16bits, and freaked out the stm32 it seems)... some more thorough investigation in what caused the Overruns would probably be useful, but disabling SPI right after starting the DMA receive seems to work.
 

### Final Note:
Infineon's datasheet mentions that the encoders have a SLAVE_NUMBER that can be assigned 0-3 so a max of 4 encoders on one bus, but in reading the datasheet it seems that's only due to the Short PWM Code (SPC) interface. In SPC, the encoders would listen for different length pulses (based on their SLAVE_NUMBER), then send out a PWM pulse with the absolute position. SPC is is different from the SPI/SSC interface, so I have ignored it. I've been able to test multiple encoders together with the same slave number, never writing to the register, and they work completely fine... SPI only uses the CS pin so in theory you could have as many as GPIO pins available.

### DIFF
--- a/Firmware/Drivers/STM32/stm32_spi_arbiter.cpp
+++ b/Firmware/Drivers/STM32/stm32_spi_arbiter.cpp
@@ -35,17 +35,39 @@ bool Stm32SpiArbiter::start() {
     if (!equals(task.config, hspi_->Init)) {
         HAL_SPI_DeInit(hspi_);
         hspi_->Init = task.config;
-        HAL_SPI_Init(hspi_);
+        HAL_SPI_Init(hspi_);  
     }
+
+    
     task.ncs_gpio.write(false);
     
     HAL_StatusTypeDef status = HAL_ERROR;
-    if (task.tx_buf && task.rx_buf) {
-        status = HAL_SPI_TransmitReceive_DMA(hspi_, (uint8_t*)task.tx_buf, task.rx_buf, task.length);
-    } else if (task.tx_buf) {
-        status = HAL_SPI_Transmit_DMA(hspi_, (uint8_t*)task.tx_buf, task.length);
-    } else if (task.rx_buf) {
-        status = HAL_SPI_Receive_DMA(hspi_, task.rx_buf, task.length);
+    if(!task.split_tx_rx) {
+        if (task.tx_buf && task.rx_buf) {
+            status = HAL_SPI_TransmitReceive_DMA(hspi_, (uint8_t*)task.tx_buf, task.rx_buf, task.length);
+        } else if (task.tx_buf) {
+            status = HAL_SPI_Transmit_DMA(hspi_, (uint8_t*)task.tx_buf, task.length);
+        } else if (task.rx_buf) {
+            status = HAL_SPI_Receive_DMA(hspi_, task.rx_buf, task.length);
+        }
+    } else {
+        if(!task.done_tx) {
+            status = HAL_SPI_Transmit_DMA(hspi_, (uint8_t*)task.tx_buf, task.length);
+        } else {
+
+            // status = HAL_SPI_Receive(hspi_, (uint8_t*)task.rx_buf, task.length, 1000);
+            // task_list_->ncs_gpio.write(true);
+            // if (task_list_->on_complete) {
+            //     (*task_list_->on_complete)(task_list_->on_complete_ctx, true);
+            // }
+            // CRITICAL_SECTION() {
+            //     task_list_ = task_list_->next;
+            // }
+
+            status = HAL_SPI_Receive_DMA(hspi_, (uint8_t*)task.rx_buf, task.length);
+            __HAL_SPI_DISABLE(hspi_);
+
+        }
     }
 
     if (status != HAL_OK) {
@@ -90,6 +112,8 @@ bool Stm32SpiArbiter::transfer(SPI_InitTypeDef config, Stm32Gpio ncs_gpio, const
         .on_complete = [](void* ctx, bool success) { *(volatile uint8_t*)ctx = success ? 1 : 0; },
         .on_complete_ctx = (void*)&result,
         .is_in_use = false,
+        .split_tx_rx = false,
+        .done_tx = false,
         .next = nullptr
     };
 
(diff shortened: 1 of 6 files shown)


PAJohnson:
Hi Jon, thanks for submitting this PR! I've got a motor + TLE5012B on my bench right now and I'm able to use the encoder with SPI comms no problem. Regarding the "disable SPI to prevent overruns" - from what I've read, in bidirectional mode, the SPI peripheral will continue trying to clock in data after the tx-rx transition until you turn it back to tx. Disabling SPI at the end of the transfer seems reasonable to me.

Firmware/Drivers/STM32/stm32_spi_arbiter.cpp
```diff
@@ -106,19 +124,44 @@ void Stm32SpiArbiter::on_complete() {
     if (!task_list_) {
         return; // this should not happen
     }
+    if(!task_list_->split_tx_rx) {
```
PAJohnson: In this section, from line 127 to 166, can you change this to be 

```if (task_list_->split_tx_rx && !task_list_->done_tx) {TLE special case} else {"normal" case}```?

File: Firmware/MotorControl/encoder.cpp [hunk 2]
```diff
@@ -32,31 +32,46 @@ bool Encoder::apply_config(ODriveIntf::MotorIntf::MotorType motor_type) {
...
+    
+    if(mode_ == MODE_SPI_ABS_TLE) {
+        spi_task_.config = {
+            .Mode = SPI_MODE_MASTER,
+            .Direction = SPI_DIRECTION_1LINE,
+            .DataSize = SPI_DATASIZE_16BIT,
+            .CLKPolarity = SPI_POLARITY_LOW,
+            .CLKPhase = SPI_PHASE_2EDGE,
+            .NSS = SPI_NSS_SOFT,
+            .BaudRatePrescaler = SPI_BAUDRATEPRESCALER_8,
```
PAJohnson, line 55: I'm slightly concerned about running at a higher clock rate due to how much trouble people have with the lower clock rate and other SPI encoders, but it works on my test bench and it seems to be beneficial for your use case so I'm just noting this down.

PAJohnson (changes requested):
See the other comments for some direct change suggestions. 

In the datasheet for this encoder, it mentions a "Safety Word" that is automatically (?) transmitted if another 16 bit word is clocked in after the angle data. Do you have any interest in that functionality? If not, I'm OK with merging this PR without the safety word.

Firmware/MotorControl/encoder.cpp
jonmendenhall: Interesting... I didn't know people were having some trouble with the low clock rates on other SPI encoders. Do you know what sort of conditions they were running the encoders in? Using plain jumper cables right next to huge Motors and high power cables is probably not a good idea... twisted pairs with ground or shielding would help I bet.

Firmware/Drivers/STM32/stm32_spi_arbiter.cpp
jonmendenhall: Sorry I've been caught up with other stuff... yep I can look into it. It sounds like it would still work with the three cases, just by merging the TLE special rx case into the normal case.

jonmendenhall: I didn't care much about the safety word for now, but it wouldn't be difficult to include support for it and connect it with the error codes for the encoder >> axis

Firmware/MotorControl/encoder.cpp
PAJohnson: This is actually a big enough issue that we're in the process of releasing a differential SPI board to handle it. It seems to depend on the person and the encoder - some people have no problems with twisted pairs of some length, other people can't get reliable results even at short lengths.

Firmware/Drivers/STM32/stm32_spi_arbiter.cpp
PAJohnson: I've done some work to get the safety word working. You can find it on the branch `pj_temp`. There's something strange happening with the DMA, but it otherwise seems to work.

PR #531, opened 2020-11-24
Hall calib

### LINKED ISSUE #500: Calibrate all Hall edges
Some hall sensors are not perfectly aligned inside the motor, and some sensors may not be 50% duty cycle.
As long as these imperfections are repeatable, we can get close to perfect performance from them anyway by calibrating the phase angle of all 6 transition edges (or 5DoF if we calibrate total phase offset later).

Some ideas for implementation:
- Spin up the motor in a lockin spin. Sample lockin phase to make a a phase table.
- Use phase table to get the expected encoder response instead of floor for the phase detector.
- We need to compensate for the fact that the phase offset calibration uses raw counts. This means that we need to shift the phase table "center" positions such that they represent the same value as the raw count (nominally).
- While we are at it we could probably make sure hall phase offset calibration uses quite long distance.

madcowswe:
Implements #500


tools/.vscode/launch.json
```diff
@@ -9,7 +9,7 @@
             "type": "python",
             "request": "launch",
             "stopOnEntry": true,
-            "pythonPath": "${command:python.pythonPath}",
+            "python": "${command:python.pythonPath}",
```
PAJohnson, line 12: I didn't make this change on purpose - seems like a vscode thing. A tooltip indicates that `pythonPath` "is not allowed" in this launch.json file but `python` is allowed.

PAJohnson (commented):
Looks good to me! Tested and working on a motor with hall effect sensors.

PR #537, opened 2020-12-03
Add interface for setting periodic outbound CAN messages

fllmn:
Needed a way to change the periodicity and type of the periodically transmitted CAN messages, from CAN. Basically something similar to transmit PDOs. 

Just noticed there's a big refactoring of the CAN components so please kill this PR it's a bad time. If it's of any interest I'm happy to redo this after the refactoring.


craftit:
I was just about to create a fork to implement something like this, it would be very useful!

samuelsadok:
Sorry for letting this sit for a while. I think this is a useful feature in general. I will add a review in a bit.

Note that there's a set of low level refactoring that I will push in the next days. So rather than asking to resolve merge conflicts now and then ending up with new merge conflicts I think it's better to wait for those changes.

fllmn:
No worries, I've spent the time testing the feature.

Sound like a good idea, I do expect quite a lot of conflicts as this was done prior to the last CAN refactoring but there's no reason to do it twice. Will keep an eye out for the refactoring commit.

fllmn:
All conflicts should be resolved by now but there still some testing to be done. I can notify when I know everything is working as intended.

Question related to the merge, the new CAN thread now implements a sleep between the `service_stack()` calls. What's the best way wake the thread to reschedule when a periodic update gets enabled? I solved it by extending the `CanBusBase` interface with a function to release the semaphore. It's possible that it's not needed at all since CAN traffic will release the semaphore in the callbacks but I'm not sure that's a reliable strategy.

samuelsadok:
Actually I just pushed the low level CAN changes to devel now, so now there are conflicts again... that was a bit unfortunate timing, sorry ^^

I tried to anticipate this PR a bit so it shouldn't be too hard to add configurable periodic messages to the refactored code. There's already an array for periodic messages defined [here](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/communication/can/can_simple.hpp#L111) and set up [here](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/communication/can/can_simple.cpp#L25). The periodic callback is implemented [here](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/communication/can/can_simple.cpp#L80).

I saw that you added a new command to dynamically start/stop selected messages via CAN. Is there a strong requirement to do so? I prefer to keep state as minimal as possible. For reference in CANopen you can, as far as I'm aware, only start/stop messages altogether and you can reconfigure them (but that's non-volatile configuration rather than volatile state). When it comes to one-time configuration the plan is to have that be done over the Fibre protocol (currently USB only, but perhaps in the future over CAN too).

Regarding the semaphore question: the new code uses a timer callback to schedule the TX handlers on an event loop. If it's not viable to wait for a timer to expire, the timer interface would need to be carved out more to allow for cancelling an ongoing timer.

fllmn:
I agree, shouldn't be to hard to resolve.

In regards to the dynamic reconfiguration from CAN. In my implementation it's a quite useful feature. I'm using an external controller for position/velocity and I need a way to get torque/position feedback at a high rate. Both channels are used but exclusively, meaning that one not in use gets disabled which in turns free up the bus. But I also understand the a feature like this goes against the idea of a "Simple CAN" interface. That's interesting regarding CANopen, I was under the opposite impression :), or at least that PDOs were configurable in pre-operational state but my experience is quite limited.

I guess it's up you if you want me to retain the dynamic rescheduling over CAN or not. I might offer it in the fork since it quite useful for me. If we exclude the dynamic rescheduling we don't have to worry about waking the thread.

samuelsadok:
Ah I see. I feel like this is a bit a niche case though. As you said this protocol is supposed to remain simple so I prefer not to integrate this level of fine grained control. I see CANSimple as a temporary solution that should create minimal maintenance burden until we have support for an established protocol like CANopen.

Yes PDOs are reconfigurable via SDOs but I think those settings are supposed to be persistent across reboots and I'm not sure if they are required to take effect immediately. The practical difference is that with persistent configuration you don't need to worry about synchronizing many state variables between nodes. You just configure a node once when building the machine and after that at every point in time each node's communication subsystem is in exactly one of four well-defined states, regardless of unexpected resets of master/slave nodes.

My knowledge about CANopen is also limited so my assumptions could be off a bit but if on-the-fly reconfiguration is a thing I would still lean on the maintenance argument mentioned above.

PR #542, opened 2020-12-21
workaround to fix deprecated .gcd function

MxFxM:
When trying to flash the newest firmware, the odrivetool crashed. It turns out there is a deprecated function .gcd used. I was able to work around it by using the function included in the math module.


madcowswe:
Thanks!

Issue #547, opened 2021-01-13
Homing not returning to zero position


fllmn:
**Describe the bug**
When running the homing procedure the motor moves until end-stop is pressed and then stops. As the end-stop is still pressed the motor will be unable to go into CLC as the end-stop will generate an error.

The underlying issue seems to be that `input_pos_` and `pos_setpoint_` will get updated to the same value when `start_closed_loop_control` gets called. Since the trajectory generator uses the diff between `input_pos_`and `pos_setpoint_` the planned time is 0 and in turn the motor never moves.

 - OS: Linux 5.9.11-3-MANJARO
 - odrivetool Version: v0.5.1.dev

This could of course only be a problem with the dev branch.

riewert:
I have the same problem on devel. I agree with everything you say. I think it may be that the `controller_.pos_estimate_linear_src_` is not being updated correctly, before this line gets called.

https://github.com/odriverobotics/ODrive/blob/devel/Firmware/MotorControl/axis.cpp#L300

Wetmelon:
What did you set for `min_endstop.config.offset`?

riewert:
I've tried both positive and negative offsets. This is what I am currently running.

```config:
  debounce_ms: 50 (uint32)
  enabled: True (bool)
  gpio_num: 5 (uint16)
  is_active_high: False (bool)
  offset: -1.0 (float)
endstop_state: False (bool)
```

I've run through it a few times in the debugger. It looks like a single call is made [here](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/MotorControl/controller.cpp#L130) between the [closed loop control starting](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/MotorControl/axis.cpp#L415) and the trap_traj being [set](https://github.com/odriverobotics/ODrive/blob/devel/Firmware/MotorControl/controller.cpp#L180). The variables are at different values during this call than they were set to, hence the trap_traj gets weird input namely that `input_pos_` and `pos_setpoint_` are equal.

riewert:
Fixed it on my system. Here is a pull-request.

Also added a small change, the offset position is set to 0 instead of the trigger point.

https://github.com/odriverobotics/ODrive/pull/550

fllmn:
Just like @riewert , the endstop configuration wasn't doing anything. I also tried negative and positive offsets. 

[...]

PR #550, opened 2021-01-27
Set input_pos after init

riewert:
This should fix the homing issues described in #547 .


Firmware/MotorControl/axis.cpp
```diff
@@ -397,29 +397,29 @@ bool Axis::run_homing() {
...
 
-    // pos_setpoint is the starting position for the trap_traj so we need to set it.
-    controller_.pos_setpoint_ = min_endstop_.config_.offset;
-    controller_.vel_setpoint_ = 0.0f;  // Change directions without decelerating
-
-    // Set our current position in encoder counts to make control more logical
-    encoder_.set_linear_count((int32_t)(controller_.pos_setpoint_ * encoder_.config_.cpr));
-
+    // Calculate the desired position after offset.
+    float inputBuffer = encoder_.pos_estimate_.any().value() + min_endstop_.config_.offset;
```
samuelsadok: We need to check if `encoder_.pos_estimate_.any()` is `std::nullopt`. If so, `.value()` shouldn't be called and we need to abort homing.
`inputBuffer` => `input_buffer`

samuelsadok (changes requested):
Thanks for the fix! Just a minor comment from my side.

Firmware/MotorControl/axis.cpp
riewert: Added an error for the abort case, and fixed my naming convention. Too much Java lately...

See latest [commit](https://github.com/odriverobotics/ODrive/pull/550/commits/d68dfbda824ffb336dccf5eb3ee2c602998c7551).

Firmware/MotorControl/axis.cpp
```diff
@@ -397,29 +397,34 @@ bool Axis::run_homing() {
...
     error_ &= ~ERROR_MIN_ENDSTOP_PRESSED; // clear this error since we deliberately drove into the endstop
 
-    // pos_setpoint is the starting position for the trap_traj so we need to set it.
-    controller_.pos_setpoint_ = min_endstop_.config_.offset;
-    controller_.vel_setpoint_ = 0.0f;  // Change directions without decelerating
-
-    // Set our current position in encoder counts to make control more logical
-    encoder_.set_linear_count((int32_t)(controller_.pos_setpoint_ * encoder_.config_.cpr));
+    
+    if (encoder_.pos_estimate_.any() == std::nullopt || !encoder_.pos_estimate_.any().has_value()){
```
samuelsadok: There's a possible race condition here: `encoder_.pos_estimate_.any()` could become `std::nullopt` between this line and the next. To fix this you can read `encoder_.pos_estimate_.any()` into a local variable and then operate on that.

riewert: Thanks for pointing that out. Modified it [here](https://github.com/odriverobotics/ODrive/pull/550/commits/7ac7c6ec58c6eabe7d832c7c9b105504c66c4242).

samuelsadok: Great, thanks! Looks good now.

samuelsadok (approved):

PR #552, opened 2021-01-31
Add build for MacOSX

tmagik:
This should be necessary, but not sufficient to build libfibre for arm64-apple-darwin20.1.0.

The documentation or scripts on how this is supposed to work are also not particularly obvious either.


tmagik:
https://github.com/samuelsadok/fibre/pull/15 looks like the right way to fix this. (it also makes more sense once I realized the macos .dylib is cross-compiled from linux)

PR #553, opened 2021-02-05
Optimizre can_getSignal and can_setSignal functions

Wetmelon:
Breaks strict aliasing rules for _maximum speed_


madcowswe:
looks good to me. Do we have any tests on target?

samuelsadok (approved):
Passes HWIL CAN tests on v3.6.

madcowswe:
Awesome!

PR #555, opened 2021-02-21
'Current control' linked updated to read 'Torque control'

Capo01:
The current control link pointed nowhere. Updated to torque control.


samuelsadok:
Thanks! The anchor `#Torque control` needs to be without spaces but I just fixed the link directly in master (75dc78c3e3a81d2201fe8bd8702d7670d66072c8)

PR #556, opened 2021-03-05
Switch from depracated fractions module to math

tmagik:
Fixes errors on MacOS big sur with python 3.9.2 from homebrew


samuelsadok:
Thanks! Looks like you need to use your GitHub email address for the License Agreement check to pass. You can edit the commit and then force push. Please also rebase/cherry-pick to the devel branch.

Issue #539, opened 2020-12-09
SPI Encoders randomly stop working after reset (nRST)


OstapFerensovych:
I am using AS5048A encoders on both axes.
I get ERROR_ABS_SPI_COM_FAIL randomly on one or both axis after I reset the STM chip by nRST pin.
`odrv0.axis0.encoder.error = 0` `odrv0.axis1.encoder.error = 0` do not clear the error. While the error is present spi generates clocks and data is present on MISO from both encoders.
After the power cycle, the errors are gone.

To reproduce this behavior connect two AS5048A encoders. (In my case I use GPIO4 and GPIO8 as nCS pins)
Reset STM32F405 by touching nRST to GND.

 - Firmware v0.5.1-release
 - ODrive HW v3.6-56V
 - OS: Ubuntu 20.10
 - odrivetool Version v0.5.1.post0

OstapFerensovych:
After some debugging, I found that AMS encoders have an error flag at the 14th bit, and ODrive firmware checks this flag, but it does nothing when this flag is `1`:
`if (ams_parity(rawVal) || ((rawVal >> 14) & 1)) { return; }`

It looks like while I do reset on the nRST pin - the SPI clock interrupts and AS5048A rises error flag.

Information from AS5048A datasheet.
> Possible conditions which force the ERROR FLAG to be set:
...
-Wrong number of clocks (no full transmission cycle or too
many clocks)
...

So is there a need to check this flag if there is a parity check implemented in firmware?

samuelsadok:
Thanks for debugging this!

Looking at the datasheet it's unclear to me what exactly happens when the encoder detects a parity error in a command coming from the master. Is the next response from the encoder a valid angle or is it all zeros? If it's a valid angle then we can indeed just ignore the error flag. If not then we need to ignore the response and somehow send a clear-fault command, otherwise we cannot detect the next time the encoder detects a parity error.

Since I don't have this encoder at hand would you mind testing this? You can flip a bit in `abs_spi_dma_tx_` in encoder.hpp and see if the encoder still returns valid angles.

OstapFerensovych:
Yes, I already tested this!
I simply removed ` || ((rawVal >> 14) & 1)` and looks like everything works fine. The angle is valid when the error flag is 1.

[...]

PR #563, opened 2021-04-21
Add ignore_abs_ams_error_flag

hbuhle2s:
This adds an option to disable the errorflag check for abs_ams encoders.
See also here:
https://discourse.odriverobotics.com/t/clarification-on-spi-encoders/6451

### DIFF
--- a/Firmware/MotorControl/encoder.cpp
+++ b/Firmware/MotorControl/encoder.cpp
@@ -562,9 +562,18 @@ void Encoder::abs_spi_cb(bool success) {
     switch (mode_) {
         case MODE_SPI_ABS_AMS: {
             uint16_t rawVal = abs_spi_dma_rx_[0];
-            // check if parity is correct (even) and error flag clear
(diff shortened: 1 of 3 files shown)


aske-p:
Hi! I'm having trouble compiling the firmware (never done it before...), but i could really use this feature of disabling the error flag. Is there any chance you could compile the 5.4 fw version for V3.6-56v to hex and commit it?

grahameth:
Hi,
here you go:
https://github.com/grahameth/ODrive_hbuhle2s_0.5.4_abs_ignore/blob/ignore_abs_ams_error_flag_0_5_4/ODriveFirmware_3.6_56V.hex
I didn't test this version though (I am personally still using the 0.5.1 branch). So let me know if it works, maybe soneone else will want to use it too.

best regards

samuelsadok:
Sorry actually for closing this without comment. Maybe it was automatic when renaming the target branch.
Anyway, for context, here's the related discussion: https://github.com/odriverobotics/ODrive/issues/539

aske-p:
@grahameth  Thanks! It does indeed work

PR #567, opened 2021-05-12
Adding test inputs and data collection for motor characterization

labmrd:
This makes it pretty quick and easy to hand-characterize the ODrive's control of a motor, i.e. get the transfer function for "the motor's position output as a function of the ODrive's voltage command", which can be used for control design (e.g. designing the ODrive controller gains rather than doing manual tuning).

How it works:
Adds a new axis state to run a voltage test input (impulse, step, exponential chirp, or white noise) using parameters stored in new axis field input_config_ (which has also been added to save_configuration() and load_configuration()). While input is running, data (timestep, input, encoder position and velocity) are continuously saved to a ring buffer (modeled on oscilloscope). User can call odrivetool method run_motor_characterize_input() to set axis state, pull data from ring buffer as fast as possible (I got around 150-200 Hz on my system), and export it to CSV. See [here](https://github.com/labmrd/odrive-with-motor-characterization/blob/main/docs/motor_characterization_references/report/Summary%20of%20Contributions_v5.pdf) for more details.

This work was completed for a master's project in the Medical Robotics and Devices Lab at the University of Minnesota, and for that purpose a static version of this material is stored [here](https://github.com/labmrd/odrive-with-motor-characterization). See [here](https://github.com/labmrd/odrive-with-motor-characterization/tree/main/docs/motor_characterization_references) for project report, presentation, and demo data.

Things that could use optimization:
- Probably better to consolidate the four ring buffer get functions into one, taking two index arguments instead of one each.
- utils method currently requires the odrive object as an argument; there's definitely a better way to do that, I just didn't know how
- Can optimize for speed if you want by reducing the data collected (e.g. deriving velocity offline rather than pulling the encoder velocity estimates, or treating the input as known and not bothering to collect data for it)

(Disclaimer that this is my first time using git for anything more complicated than personal file backup, so I made some rookie mistakes with forking and merging, thus all of the "cleaning up after mergetool" commits. I think it's good now, but apologies in advance if I missed any pull request etiquette details.)


Wetmelon:
Hmm, nobody ever responded?  This is a good PR, but it needs to be updated. 
Please use the `INPUT_MODE` system instead of unique axis states - we added a "TUNING" mode, which is really sine-waves.  Other Tuning styles (chirps, etc) could be added.

I'm going to close this PR for now, please reopen if you still desire to merge branches :)

PR #583, opened 2021-06-14
Arduino improvements

aarondls:
New autogenerated ODriveEnums.h header file to replace old hardcoded inline enum in ODriveArduino class
- Created the arduino_enums_template Jinja template for autogenerating the Arduino enums header file.
- Added Python autogeneration command for the header file Arduino/ODriveArduino/ODriveEnums.h to Makefile. Reflected the change in the Dockerfile and also added note to update ODriveEnums.h along with enums.py in developer-guide markdown file 
- Replaced old inline enum in the ODriveArduino class by including the autogenerated enum header file ODriveEnums.h in ODriveArduino.h. This reduces the possibility of the enum being out of date with the current API (as it is right now), since it does not have to be manually updated and checked every time.
- Removed the scope resolution operator to access the enums as it can now be accessed from the global namespace.

New GetPosition member function
- Added GetPosition member function to ODriveArduino class to complement existing GetVelocity, SetVelocity, and SetPosition member functions.
- Changed the example ODriveArduinoTest.ino file to reflect the new GetPosition member function compared to the old verbose serial command.


Wetmelon (commented):
Looks good to me.  Very thorough.

samuelsadok:
Thanks for the work, looks good to me!

PR #584, opened 2021-06-14
Update firmware version prerelease status resolution

clintlombard:
Update `odrive.dfu.update_device` firmware version prerelease status resolution to use `version_unreleased` attribute, as opposed to using the `version_prerelease` attribute (no longer a valid attribute). Backwards compatibility is ensured.


samuelsadok:
Good catch. This variable was actually always called `fw_version_unreleased` and not `fw_version_prerelease` so you can remove the `if` logic. Note that I changed the target to the `fw-v0.5.3` branch so please rebase to that branch. Also please make sure that there are no spurious whitespace changes.

clintlombard:
Sure thing. Done :+1:

PR #592, opened 2021-07-08
Update can-protocol.md

m84u:
Set the baud rate with ```set_baud_rate(500000)``` instead of ```baud_rate = 500000```

### DIFF
--- a/docs/can-protocol.md
+++ b/docs/can-protocol.md
@@ -22,7 +22,7 @@ Each axis looks like a separate node on the bus. Thus, they both have the two pr
 ```
 odrv0.axis0.config.can_node_id = 3
 odrv0.axis1.config.can_node_id = 1
-odrv0.can.config.baud_rate = 500000
+odrv0.can.set_baud_rate(500000)
 odrv0.save_configuration()
 odrv0.reboot()
 ```


samuelsadok:
What motivates this change? Are you using the latest firmware?
`set_baud_rate` is not an exposed function according to the yaml: https://github.com/odriverobotics/ODrive/blob/master/Firmware/odrive-interface.yaml#L390

m84u:
Setting the baud rate like in the current documentation didn't work for me. Found the working part in [CAN Interface Available for Testing](https://discourse.odriverobotics.com/t/can-interface-available-for-testing/1448/158). My version is v0.5.2.post0.

I'm not an yaml expert, but how i would read the declaration is: (c_)setter is 'set_baud_rate'.
        line390:  ```baud_rate: {type: uint32, c_setter: 'set_baud_rate'}```

samuelsadok:
What's your firmware version (`(odrv0.fw_version_major, odrv0.fw_version_minor, odrv0.fw_version_revision)`)?

There exists a firmware-internal C++ function called `set_baud_rate`, but it is exposed as a read-write property called `baud_rate`. `c_setter` is a hint for the code-generator for when compiling the firmware to ensure the correct write hook is called.

m84u:
Ok, i see. If it's jus an issue on my configuration, then please just reject this pull request. But i really thought it's a deprecated documentation and other users also would have the same confusion with that.
My fw-version is: 0.5.1 
grafik

PR #624, opened 2021-09-14
Fixed generation of version.c file to please the docker build

pbrier:
Docker was not building anymore. Failing with this error:
```
tup error: Explicitly named file 'version.c' not found in subdir 'autogen'
tup error [string "builtin"]:89: Error parsing input list
stack traceback:
	[C]: in field 'definerule'
	[string "builtin"]:89: in field 'frule'
	[string "Tupfile.lua"]:44: in function 'compile'
	[string "Tupfile.lua"]:433: in main chunk
```
Added generation of this file to the build commands. 
Just out of curiosity: Why are we not using ```make``` in the docker build??

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ WORKDIR ODrive/Firmware
 # Must attach the firmware tree into the container
 CMD \
 	# Regenerate python interface
+	mkdir ../Firmware/autogen && \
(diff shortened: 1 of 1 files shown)


pbrier:
Dropping this PR: Need to create PR with other account. Also: this fix only builds a clean tree, as the` autogen` directory may not exist before the build

PR #626, opened 2021-09-16
Added integrator_limit to allow prevention of integrator windup

pbrier:
In some cases the speed controller integrator can windup (if the motor is pushed out of position for a long time, or an excessive command is given), as a result the motor will overshoot significantly. This can be prevented by limiting the integrator. This PR adds a setting to the speed controller to allow you to do this. Set to `inf` (default value) to disable it and have the original behavior.

### DIFF
--- a/Firmware/MotorControl/controller.cpp
+++ b/Firmware/MotorControl/controller.cpp
@@ -364,6 +364,8 @@ bool Controller::update() {
         } else {
             vel_integrator_torque_ += ((vel_integrator_gain * gain_scheduling_multiplier) * current_meas_period) * v_err;
         }
+        // integrator limiting to prevent windup 
+        vel_integrator_torque_ = std::clamp(vel_integrator_torque_, -config_.vel_integrator_limit, config_.vel_integrator_limit);
     }
 
     float ideal_electrical_power = 0.0f;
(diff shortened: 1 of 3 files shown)


Wetmelon:
The vel_integrator actually has anti-windup in the form of a clamp and decay.  The output of the velocity loop is clamped to the current limit, and then the integrator is decayed by 1% every cycle.

Is there a particular benefit to clamping just the integrator to a particular value?

Also, please target fw-v0.5.4

pbrier:
Yes. This ensures that the integral part of the controller never exceeds a preset limit. for example: If you set it to 20% of the maximum current, a prolonged standstill with residual error would not have an output of this 20%, ever. If the external force that caused this residual error (a hard stop for example) is removed (by reversing for example), the axis would not under/overshoot excessively.

pbrier:
To be more exact: When the integrator term is more than 1% continuously, windup of the integrator will still occur (the code does mention this ```// TODO make decayfactor configurable```) but even then, a decay factor does not have the same effect as a simple clamping of the integrator. There are different techniques to prevent windup, and they all have there pros and cons. A hard clamp is the most straightforward and is implemented in many motion controllers. Some background: https://e2e.ti.com/blogs_/b/industrial_strength/posts/teaching-your-pi-controller-to-behave-part-vii

madcowswe:
There certainly is scope to limit/clamp the integrator value at some earlier limit than the configured current/torque limit, which is what the existing anti-windup does. Thanks for your contribution.

PR #640, opened 2021-11-04
Add version.c generation to dockerbuild and actions

AzazKamaz:
Actually this is fix to c5720564d74b47359a2dab45aa315ab6c0550a0e that have broken `dockerbuild.sh` and seems to broke GitHub Action.

Reason is that c5720564d74b47359a2dab45aa315ab6c0550a0e removes generating `autogen/version.c` from `Tupfile.lua` and adds it to `Makefile`. But `dockerbuild.sh` and GitHub Action does not use `Makefile` so they can't find `autogen/version.c`

If `dockerbuild.sh` or GitHub Action works I think the reason is that they have old `autogen/version.c` in cache. Don't know about GitHub Action but `dockerbuild.sh` doesn't work at this moment from clear repo

### DIFF
--- a/.github/actions/release-firmware/action.yml
+++ b/.github/actions/release-firmware/action.yml
@@ -19,6 +19,9 @@ runs:
 
         cd ${{ github.workspace }}/Firmware
 
+        mkdir -p autogen
+        python ../tools/odrive/version.py --output autogen/version.c
+
         echo "CONFIG_STRICT=true" > tup.config
         echo "CONFIG_BOARD_VERSION=${{ inputs.board_version }}" >> tup.config
         tup init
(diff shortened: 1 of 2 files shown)


AzazKamaz:
The right way to regenerate `autogen/version.c` each time is to tell tup to run generating task (it was commented in c5720564d74b47359a2dab45aa315ab6c0550a0e) always (not only when there is no output file). But it doesn't have such option ;)

Another option is to delete `autogen/version.c` each time before running tup build. It is better way than current but I think it is a workaround too. So I have just added generating `autogen/version.c` to other places where tup is called (like in original c5720564d74b47359a2dab45aa315ab6c0550a0e)

samuelsadok:
Thanks for the fix.

The root problem here is actually that tup doesn't monitor hidden files, thereby breaking its soundness guarantees in cases like this one, where `version.c` depends on the hidden git index. A generating task would therefore only fix the symptom but not the root problem.

PR #641, opened 2021-11-04
Fix race condition in trap traj control

AzazKamaz:
Actually there is a race condition at [controller.cpp:212](https://github.com/odriverobotics/ODrive/blob/68d923e8a8b60e21308cba41a66d0ca18c84dd91/Firmware/MotorControl/controller.cpp#L212):
1. When execution is somewhere in lines 206-211 this function can be interrupted (for example by CAN Bus) and `input_pos_` can be updated.
2. Then it will be immediately set into the `pos_setpoint_` and motor will go there by PID instead of trap traj.
3. On the next iteration `move_to_pos(input_pos_)` will plan next trajectory from `pos_setpoint_` and immediately complete it because `pos_setpoint_ == goal_point` so motor will continue moving by PID

To catch this case this part of code can be commented, then problem line will be executed in each iteration when trajectory have been completed and it is very easy to catch the bug:
```
if (trajectory_done_)
    break;
```


Wetmelon (approved):

Wetmelon:
Good catch, thanks!

PR #652, opened 2022-01-11
Adds new brake axis state and an optional fallthrough on watchdog timeout

Jonas-Witt:
This adds a new axis state `AXIS_STATE_BRAKE` that is the same as closed loop control with forced velocity control & zero velocity input. The main purpose of this new state is to allow a defined axis state that we can safely fallback to in case of a watchdog timeout. A new axis configuration `enable_brake_state_fallthrough_on_watchdog_timeout` controls this behavior. This fallthrough is a useful safety feature for wheeled robots that may encounter a state where the controlling node disappears and a watchdog timeout is triggered to keep the wheeled robot from rolling down an incline uncontrolled in idle mode.


Wetmelon:
Hi Jonas, thanks for the PR.  Do you have test data you can share?

Firmware/MotorControl/axis.cpp
```diff
@@ -169,16 +169,30 @@ void Axis::watchdog_feed() {
...
-    } else {
-        error_ |= ERROR_WATCHDOG_TIMER_EXPIRED;
-        return false;
     }
+    // Go to braking state from closed loop control when watchdog times out.
+    if (current_state_ == AXIS_STATE_CLOSED_LOOP_CONTROL &&
+        config_.enable_brake_state_fallthrough_on_watchdog_timeout) {
+        //controller_.config_.control_mode = Controller::CONTROL_MODE_VELOCITY_CONTROL;
+        //controller_.config_.input_mode = Controller::INPUT_MODE_PASSTHROUGH;
+        //controller_.input_vel_ = 0;
```
Wetmelon, line 185: Please remove commented code

Wetmelon:
Is there anything in this PR to prevent spurious communications from fighting with the watchdog for control of the axis?

Firmware/MotorControl/axis.cpp
```diff
@@ -362,6 +376,22 @@ bool Axis::run_closed_loop_control_loop() {
     return check_for_errors();
 }
 
+bool Axis::run_brake_loop() {
+    start_closed_loop_control();
+    set_step_dir_active(config_.enable_step_dir);
+
+    while ((requested_state_ == AXIS_STATE_UNDEFINED) && motor_.is_armed_) {
+        controller_.config_.control_mode = Controller::CONTROL_MODE_VELOCITY_CONTROL;
+        controller_.input_vel_ = 0.0f;
```
Wetmelon, line 385: What if the axis isn't in in `INPUT_MODE_PASSTHROUGH`?

madcowswe:
I would say it's better to avoid a new state that is the same as closed loop control just with some overrides in the inputs. I would suggest to instead just override the inputs directly right before the "update inputs" section of controller.cpp
image

Wetmelon:
This PR should be refactored and made to fix #704

PR #664, opened 2022-02-09
Fix axis_state signal length

Wetmelon:


nick-a-schneider:
I can't explain why the macOS build failed to install arm-none-eabi-gcc, although I have noticed significant delays while installing the dependencies for these tests on the past few PRs... I'll bring it up with Sam. Anyways, re-running the tests and they all passed. Unless you have any other concerns you can ahead and merge, or I can do it in the morning if you haven't already.

Issue #666, opened 2022-02-12
Some enum types use values that don't fit in the type


Wetmelon:
**Describe the bug**
Some enums (e.g. Motor::Error) have values that are too large for the underlying type.  All values above 2^32 are replaced with 0 (e.g. `Motor::Error::NONE`)

For Motor::Error, this is:

- ERROR_UNKNOWN_VOLTAGE_COMMAND
- ERROR_UNKNOWN_GAINS
- ERROR_CONTROLLER_INITIALIZING
- ERROR_UNBALANCED_PHASES

**Additional context**
https://discourse.odriverobotics.com/t/bug-with-motor-error-enums/8685/2

PR #667, opened 2022-02-12
Make all autogenerated enums int64_t

Wetmelon:
Quick fix for #666

### DIFF
--- a/Firmware/fibre-cpp/interfaces_template.j2
+++ b/Firmware/fibre-cpp/interfaces_template.j2
@@ -36,7 +36,7 @@ public:
 [[render_interface(intf) | indent(4)]]
 [%- endfor %]
 [%- for enum in intf.enums %]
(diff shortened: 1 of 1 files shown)


Wetmelon:
> The only exception are various versions of MSVC, which unreasonably insist on using a 32 bit type even though it's too small.

Ah you know what?  I was accidentally using MSVC when I checked `std::underlying_type_t` against `Motor::Error`.  `arm-none-eabi-gcc` 10.2 shows `unsigned long long`...  godbolt was a better choice.

Nevertheless, setting it explicitly feels safer

samuelsadok:
The problem with this is that it makes all enums unnecessarily RAM-hungry, even the ones like AxisState that would fit into one byte. So I think if we want to solve this (which is not super high prio if it's only an issue on MSVC) we should solve it dynamically.

PR #671, opened 2022-02-15
Two Dockerfile fixes

szeder:
The first commit fixes the build with the `Dockerfile`, the second avoids a potential caching issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,6 @@ CMD \
 	python ../tools/odrive/version.py \
 	--output autogen/version.c && \
 	# Regenerate python interface
-	mkdir ../Firmware/autogen; \
 	python interface_generator_stub.py \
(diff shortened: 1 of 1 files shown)


Wetmelon (approved):

szeder:
Force-pushed to correct two typos in the second commit's log message, no other changes.

samuelsadok:
Thanks, can you rebase the branch onto `devel`?

szeder:
I've rebased it on top of the `devel` branch.  It turned out that `devel` already contained a fix for the missing `version.c` issue, making my first commit unnecessary, so I dropped it.
However, it turned out that that issue has been fixed _twice_ in `devel`, and `version.c` is now generated twice during a containerized build.  So I added a new commit to remove the redundant commands.

PR #672, opened 2022-02-17
Turn controller into PID controller

amakuzaze:
Hi,

this code was written by a service contractor whom we have gotten written permission to upstream it in their name.
I personally am not a mechanical engineer, so I can only answer questions in a limited capacity, but I'll do my best.

Thanks

### DIFF
--- a/Firmware/MotorControl/controller.cpp
+++ b/Firmware/MotorControl/controller.cpp
@@ -328,6 +328,11 @@ bool Controller::update() {
         v_err = vel_des - *vel_estimate;
         torque += (vel_gain * gain_scheduling_multiplier) * v_err;
 
+        float vel_error_diff = v_err - vel_err_previous_;
+        vel_error_diff /= current_meas_period;
+        vel_err_previous_ = v_err;
+        torque += vel_error_diff * config_.vel_differentiator_gain;
+
         // Velocity integral action before limiting
         torque += vel_integrator_torque_;
     }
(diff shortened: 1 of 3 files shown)


szeder:
Please update the commit message to mention the right PID term: it claims that it adds the "P part", but in fact it adds the derivative term.

amakuzaze:
@szeder thanks. also rebased onto master / resolved conflicts while I was at it.

samuelsadok:
@amakuzaze could you provide some more insight into the reason the D-term was needed?

Theoretically, acceleration can be changed instantaneously*, so it seems there would be no D-term needed for velocity control. However maybe your contractor was facing a special case where our model doesn't match the reality.

Maybe it was done to compensate for a lag in the encoder velocity estimate. In this case it might be better though to adapt the encoder estimator accordingly (e.g. take the control output into account).

(*) assuming the usual case where the electrical dynamics are much faster than the mechanical dynamics

samuelsadok:
Closing due to inactivity. Let me know if any new information turns up.

szeder:
FWIW, in a recent project I wanted to use this derivative term to reduce overshoot on velocity setpoint changes (our robot was somewhat heavy), and even built and flashed a firmware with this PR's changes included, but never got around to actually tune it.

In retrospect, I think I could have somehow used the torque feed-forward to the same effect, but its documentation is unclear and sparse, so I'm not sure how to use it, or whether that's suitable for the task in the first place. (and that project was shot down anyway).

PR #682, opened 2022-03-30
Add back subcomponent error flags

samuelsadok:
These flags were previously removed in https://github.com/odriverobotics/ODrive/commit/5661a3b5ed93591d3bfcd81dc355e8cd4079e2d7


szeder:
The commit reference included in the log message points to a non-existing commit, because the object name is missing its first hexdigit; it should be 5661a3b5 (Refactor control loop., 2020-09-01).

Anyway, as I was recently considering configuring an `error_gpio_pin` and hook it up to the microcontroller controlling the ODrive, I think bringing back subcomponent error flags is a good idea.

madcowswe (approved):

samuelsadok:
@szeder you're right, thanks! Fixed now.

PR #684, opened 2022-03-31
changed CANSimple::get_iq_callback to set iq_setpoint and iq_measured…

dylanmeehan:
… (as described in the docs: https://docs.odriverobotics.com/v/latest/can-protocol.html). before, this function incorrectly set id setpoint and iq setpoint


Wetmelon:
Thanks for the PR.  I've updated the devel branch to match the ODrive Pro code instead (same functional change, slightly different code)

PR #687, opened 2022-04-13
`ignore_illegal_hall_state` consistently ignores it

Wetmelon:


Firmware/MotorControl/encoder.cpp
```diff
@@ -678,7 +678,7 @@ bool Encoder::update() {
                                 edge_idx = hall_cnt;
                             } else if (mod_hall_cnt == 5) { // counted down
                                 edge_idx = last_hall_cnt_.value();
-                            } else {
+                            } else if(!config_.ignore_illegal_hall_state){
```
samuelsadok, line 681: This will lead to `edge_idx` being uninitialized when it's used further down the function.

Firmware/MotorControl/encoder.cpp
```diff
@@ -257,7 +257,7 @@ bool Encoder::run_hall_polarity_calibration() {
             hall_polarity = 0b010;
         } else if (flip_detect(states, 3)) {
             hall_polarity = 0b100;
-        } else {
+        } else if (!config_.ignore_illegal_hall_state){
```
samuelsadok, line 260: Is this fallback to (the seemingly arbitrary) `hall_polarity = 0` really what we want? Maybe we can instead just change `uint8_t states = state_seen.to_ulong();` to `uint8_t states = state_confirmed.to_ulong();`.

madcowswe: Yes polarity 0 is the "standard" polarity. There are lots of bug reports of the hall polarity detection and I think we should just give up on it on the old FW and fix it for the new one.

PR #689, opened 2022-05-05
Steps and inputmode

madcowswe:
* Sync `steps_` with input pos.
* Trigger reset of input_pos and pos_setpoint to estimate when changing control mode into position control

### DIFF
--- a/Firmware/MotorControl/controller.hpp
+++ b/Firmware/MotorControl/controller.hpp
@@ -58,6 +58,7 @@ public:
         Controller* parent;
         void set_input_filter_bandwidth(float value) { input_filter_bandwidth = value; parent->update_filter_gains(); }
         void set_steps_per_circular_range(uint32_t value) { steps_per_circular_range = value > 0 ? value : steps_per_circular_range; }
+        void set_control_mode(ControlMode value) { control_mode = value; parent->control_mode_updated(); }
     };
 
     
--- a/Firmware/MotorControl/controller.hpp
+++ b/Firmware/MotorControl/controller.hpp
@@ -81,6 +81,7 @@ public:
     bool anticogging_calibration(float pos_estimate, float vel_estimate);
 
     void set_input_pos_and_steps(float pos);
+    bool set_setpoint_to_estimate();
 
     void update_filter_gains();
     bool update();
@@ -124,7 +125,7 @@ public:
     OutputPort<float> torque_output_ = 0.0f;
 
     // custom setters
-    void set_input_pos(float value) { input_pos_ = value; input_pos_updated(); }
+    void set_input_pos(float value) { set_input_pos_and_steps(value); input_pos_updated(); }
 };
 
 #endif // __CONTROLLER_HPP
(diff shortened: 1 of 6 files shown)


tobbelobb:
TL;DR. It works.

I have tested the following things on this branch:
 - Compile and upload
 - Configured like [this](https://gitlab.com/tobben/hangprinter/-/blob/ce3fb7b099e2537346b347611d836faab74703a8/firmware/ODrive/odrive-config-AB.json) and then [this](https://gitlab.com/tobben/hangprinter-org/-/blob/version_4_newer_ODriveFirmware/doc/v4/index.html#L517)
  - Checked `input_pos` on both axes immediately after startup.
  
For axis0, the `input_pos` was 0 as expected. For axis1 I found `input_pos` to be 0.94. As if the motor had been stepped 376 times forward during startup. (Steps per revolution is 400, and 376/400 = 0.94.) I heard that this rotation also happened physically. Immediately after its index search, one of the motors made a ~1 rotation movement.

I have had these sudden movements upon startup with v0.5.1 as well, ca every second time, so it's nothing new. I've gotten so used to it that I almost didn't notice. I don't think it's related to this PR.

So I restarted the experiment:
 - `odrv0.reboot()`
 - This time, both axes' `input_pos` were 0 after startup.
 - Ran both motors back and forth a couple of times, both by setting `input_pos` inside `odrivetool` and by sending step signals from the Duet. Both behaved as expected.
  - Put both motors in torque mode via CAN.
  - Turned both motors a few turns by hand.
  - Set both motors back into position mode via CAN. They both stood still after receiving the CAN command, and entered pos mode as expected.
   - Confirmed via odrivetool that `input_pos` of both motors had been updated.

Firmware/MotorControl/controller.hpp
```diff
@@ -58,6 +58,7 @@ class Controller : public ODriveIntf::ControllerIntf {
         Controller* parent;
         void set_input_filter_bandwidth(float value) { input_filter_bandwidth = value; parent->update_filter_gains(); }
         void set_steps_per_circular_range(uint32_t value) { steps_per_circular_range = value > 0 ? value : steps_per_circular_range; }
+        void set_control_mode(ControlMode value) { control_mode = value; parent->control_mode_updated(); }
```
samuelsadok, line 61: This sets the input position to the current position estimate every time the control mode is switched. Is this important for some use cases?
I could imagine cases where a user wants to seamlessly transition from velocity control to position control with a preconfigured setpoint.

Wetmelon: Hmm, seems like a good "Law of Least Surprise" thing anyway.  Switching from vel to pos control could result in the motor suddenly moving, which isn't good for safety.  Unless the user explicitly wants that behaviour.

tobbelobb: The seamless transition into a standstill position mode, and the safety it provides are both important features for the  Hangprinter use case.

madcowswe: Going from not-position control into position control: setting the setpoint to the estimate is fine here. I would say it's better to do this because it's safer, and less surprising like Paul said. The person in your case will have to set the pos setpoint after changing mode, which I think is acceptable.

PR #696, opened 2022-06-03
Update docs on sensorless setup

jerzydziewierz:
As I wait for my encoders to arrive, I was struggling to set up the sensorless mode.

This docs update contains the lessons learned.

Plus, 
* minor grammar updates
* Python code now uses classic 4 spaces instead of 3 or 5


jerzydziewierz:
will re-commit with correct email

PR #699, opened 2022-06-23
Added get_adc_voltage()

Malaphor:
Added new enum and prototype to can_simple.hpp
Added a new entry in the switch statement and the callback function to can_simple.cpp


Firmware/communication/can/can_simple.hpp
```diff
@@ -35,6 +35,8 @@ class CANSimple {
         MSG_SET_LINEAR_COUNT,
         MSG_SET_POS_GAIN,
         MSG_SET_VEL_GAINS,
+        MSG_GET_ADC_VOLTAGE,
+        MSG_SEND_ADC_VOLTAGE,
```
samuelsadok: this value is not used anywhere

samuelsadok (commented):
Thanks!
Can you:
 - Give a brief statement what your use case is for adding this CAN message
 - Rebase onto `devel`
 - Agree to the CLA

Firmware/communication/can/can_simple.hpp
Malaphor: My use case is reading voltage from a potentiometer that measures absolute output position of a robotic joint to be used when homing or detecting minimum and maximum joint angles

I updated my files to remove the extra MSG_SEND_ADC_VOLTAGE enum so ill update the pull request to reflect that. I only use MSG_GET_ADC_VOLTAGE for both sending and receiving data

Signing CLA tonight

Malaphor: Okay, just realized I have no idea how to update the pull request. Have to work on it tomorrow

samuelsadok: You can do a fresh checkout of the ~`master`~ `devel` branch and then cherry-pick your commit on top of that and then force-push.

**edit:** master=>devel (since you're merging to devel)

Malaphor: Hmm, comparing the two they look the same. MSG_GET_ADC_VOLTAGE is used to kick off the correct callback function and MSG_SEND_ADC_VOLTAGE is used to set the txmsg.id inside the callback so the return message has a different id than the send message. So both the variables are used. If you think having a single enum would be better I can change it.

This is my first PR and I haven't used GitHub since so sorry about the confusion.

samuelsadok: On which line is MSG_SEND_ADC_VOLTAGE used? I don't see it anywhere in the code.

Regarding the git stuff, just make sure the number X in the tab "Commits (X)" shows "1", so your branch is based on our `devel` plus all your changes in one single commit. See my previous message on how to achieve that. But if it sounds too complicated could also just do a new clone and copy over the changes.

Firmware/communication/can/can_simple.cpp
```diff
@@ -343,6 +346,24 @@ bool CANSimple::get_vbus_voltage_callback(const Axis& axis) {
     return canbus_->send_message(txmsg);
 }
 
+bool CANSimple::get_adc_voltage_callback(const Axis& axis, const can_Message_t& msg) {
+    can_Message_t txmsg;
+
+    txmsg.id = axis.config_.can.node_id << NUM_CMD_ID_BITS;
+    txmsg.id += MSG_SEND_ADC_VOLTAGE;
```
Malaphor: Here I am

Firmware/communication/can/can_simple.hpp
Malaphor: On line 353 of can_simple.cpp. I left a comment above

Firmware/communication/can/can_simple.cpp
samuelsadok: Ah this was changed in a later commit so I didn't see it. Yeah this should be `MSG_GET_ADC_VOLTAGE`.

PR #707, opened 2022-08-19
Add get_adc_voltage()

Malaphor:


Firmware/communication/can/can_simple.hpp
```diff
@@ -62,6 +63,8 @@ class CANSimple {
     bool get_iq_callback(const Axis& axis);
     bool get_sensorless_estimates_callback(const Axis& axis);
     bool get_vbus_voltage_callback(const Axis& axis);
+    // msg.rtr bit must NOT be set
```
Wetmelon, line 66: What is this comment?

Malaphor: Usually when sending a message that requires a response you have to set the RTR bit but since I have to send along a gpio number I can't have the RTR bit set

Wetmelon (approved):

PR #709, opened 2022-09-06
Add controller error message

Wetmelon:
This message had a declaration but was apparently never defined.


madcowswe:
Thanks! Looks good in general. Did you test it?

Wetmelon:
Confirmed as working by @Malaphor
image

PR #710, opened 2022-09-06
Fix race condition in homing behaviour

Wetmelon:
- Sets `input_pos` and `pos_setpoint` = 0.0 after moving to the final homed position
- Add a 1ms delay to after `set_linear_count(0)` to allow `pos_estimate` to update so that we don't reset `input_pos` when transitioning back into closed loop

- [ ] Add to change log

Tested by user with industrial robot with optical homing switches.
image

image

### DIFF
--- a/Firmware/MotorControl/axis.cpp
+++ b/Firmware/MotorControl/axis.cpp
@@ -354,9 +354,6 @@ bool Axis::run_closed_loop_control_loop() {
 // Slowly drive in the negative direction at homing_speed until the min endstop is pressed
 // When pressed, set the linear count to the offset (default 0), and then go to position 0
 bool Axis::run_homing() {
-    Controller::ControlMode stored_control_mode = controller_.config_.control_mode;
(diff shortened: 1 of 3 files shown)


Firmware/MotorControl/axis.cpp
```diff
@@ -424,13 +421,16 @@ bool Axis::run_homing() {
...
+    encoder_.set_linear_count(0); 
+    controller_.input_pos_ = 0.0f;
+    controller_.pos_setpoint_ = 0.0f;
+    controller_.vel_setpoint_ = 0.0f;
     controller_.input_pos_updated();
 
-    controller_.config_.control_mode = stored_control_mode;
-    controller_.config_.input_mode = stored_input_mode;
+    // Force encoder estimate to update
+    osDelay(1);
```
samuelsadok, line 432: We have `wait_for_control_iteration()` which would be fitting here

Wetmelon: Since this is legacy homing and it was tested with osDelay, I think I'd rather just keep it

PR #720, opened 2022-09-26
Don't move count_in_cpr without moving shadow_count

Wetmelon:


samuelsadok:
I'm currently too detached from this part of the codebase to quickly judge if this change is good. Is it tested? Please also add a note to the changelog.

Wetmelon:
It's not ready to merge (not tested).  We originally added this when we were having some issues with absolute encoder async, but we ended up adding the latched value at the top of the function, so this should be redundant (but potentially troublesome).

I'm on the fence about changing it at all tbh.

PR #742, opened 2024-02-21
Hoverboard guide clarification

haroldina:
Added more descriptive text to the link to the hoverboard guide in Getting Started.

Added text above hall sensor configuration to indicate not to use that section if using an encoder added to the motor instead of the hall sensors.
Added headings for "Hall Sensor Configuration" and "Calibration and Powering the Motor"

All to clarify these steps will not apply to users of hoverboard motors with added encoders, as is common for DIY force feedback controls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
+## Important Note
+
+The firmware in this repository is compatible with the ODrive v3.x (NRND) and is no longer under active development.
+
+Firmware for the new generation of ODrives ([ODrive Pro](https://odriverobotics.com/shop/odrive-pro), [S1](https://odriverobotics.com/shop/odrive-s1), [Micro](https://odriverobotics.com/shop/odrive-micro), etc.) is currently being actively maintained and developed, however its source code is currently not publicly available. Access may be available under NDA, please [reach out to us](mailto:info@odriverobotics.com) for inquiries.
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


docs/hoverboard.rst
```diff
@@ -97,7 +97,9 @@ If you can't perform this test, a typical value is 16.
   odrv0.axis0.motor.config.current_control_bandwidth = 100
   odrv0.axis0.motor.config.torque_constant = 8.27 / <measured KV>
 
-
+Hall Sensor Configuration
+-------------------------------------------------------------------------------
+If using a hoverboard motor with an encoder attached instead of the internal hall sensors, use the encoder configuration section of the `Getting Started <https://docs.odriverobotics.com/v/0.5.6/getting-started.html#encoder-configuration>`__ guide.
```
samuelsadok: It's preferred to use `:ref:`-style references instead of hard-links (you can see the lines below for an example).

.vs/ODrive/FileContentIndex/read.lock
samuelsadok, line 1: There are several spurious files in this PR. Could you clean these up, including the changes to .gitignore (because it's not directly related to the PR). You can then use rebase to squash everything into one commit.

samuelsadok (commented):
Docs changes need to go onto the `docs-v0.5.6` branch in order to go live. Can you change the base branch of the PR and rebase onto that branch (so the diff is only one commit)?

PR #757, opened 2025-12-02
Update Dockerfile with missing package for build 'debconf'.

SiarheiHerasiuta:
This additional package needed to pass container build. Please merge this small fix with master.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ FROM ubuntu:bionic
 RUN \
 	set -x && \
 	apt-get update && \
+	apt-get -y install debconf && \
 	apt-get -y install software-properties-common && \
 	add-apt-repository ppa:team-gcc-arm-embedded/ppa && \
 	add-apt-repository ppa:jonathonf/tup && \


samuelsadok:
What happens if `debconf` is omitted?

And please note that for us to be able to consider the contribution you would need to sign the Contributor License Agreement (see comment above).

SiarheiHerasiuta:
1. - 'debconf' is needed to build firmware in Docker container way. So it is missing dependency.
2 -
image
I agree with license/cla, something wrong with github and it is not show that I am agree.

samuelsadok:
1. Can you share the error message that happens when trying to build without debconf? Sorry to be pedantic, but this is useful for documentation and future maintenance, if any.
2. Ah all good, I can see your CLA agreement on the list. Seems to be a hickup with the CLA Assistant.

SiarheiHerasiuta:
I retest master branch today on a clear Ubuntu 16.x and find just red warning (not docker build blocking as I see before and it is strange). I provide screen for better understanding.
image

samuelsadok:
I see, so it looks like a false alarm? Like the red `debconf` message presumably made you think that `debconf` is missing, but upon closer inspection it's just a warning that was emitted by `debconf` itself. If that's the case we can close this PR.

SiarheiHerasiuta:
I also test master against 'newest' Ubuntu 20, 22 and 24 and see no debconf warning message. So yes, I agree to close pull request.

PR #761, opened 2025-12-20
Add options 'deadzone' for endpoints and 'max_calib_koef' for motor

konelav:
The option 'deadzone' consists of three new attributes of ODrive.Endpoint: deadzone_min, deadzone_max, deadzone_value.
It is useful when PWM/ADC signal is used for e.g. set velocity, and can't be set precisely to get motor stopped (usually exactly middle of full range). In this case, one can add attribute values for corresponding endpoint as deadzone_min=0.4, deadzone_max=0.6, deadzone_value=0.0, so when input signal (fraction) is between 0.4 and 0.6, then output to endpoint will be exactly zero.

The option 'max_dc_calib_koef' make configurable (now hardcoded) value 0.1f -- a threshold for current measurements to be accounted as correct. It can be useful in rare cases of bad quality and/or compatibility of hardware parts, in which cases slight extending of threshold can be helpful. Or, maybe one might want to decrease threshold to have more failure-awared configuration.


samuelsadok:
Note that this repo is for ODrive v3.x and we're no longer actively working on it. We may be able to take small changes if they provide strong value and are thoroughly tested by the contributor. With that in mind:
- Deadzone/deadband is indeed a useful feature and we have just recently added this feature to our current-gen firmware (https://docs.odriverobotics.com/v/devel/manual/pwm-input.html#deadband), where it's currently available in the pre-release (`devel`). If this feature is to be backported to ODrive v3.x, it would have to follow the same API (see docs).
- The configurable `max_dc_calib_koef` seems to be for custom hardware, in which case it should be easy and sufficient to customize the value at compile time. Aside from that it should be noted that this repo does not aspire to support custom hardware.

felhub:
Very nice addition, I tested it on my v3.5-24V and it works great! Thank you!

PR #763, opened 2026-04-04
Ci v3.6 56 v

beihia:


samuelsadok:
Is this is a work-in-progress branch that was PR'd by accident?
Note that ODrive v3.6 is legacy, and we do not have the resources to take in any non-trivial changes.

